PR #9, opened 2018-04-02
Add methods for writing and reading proto binary files

val-fed:


common/FileUtils.java
```diff
@@ -128,4 +130,34 @@ public static String readFileUnchecked(String path) {
       throw new RuntimeException(e);
     }
   }
-}
+
+  /** Writes a proto to proto binary file. */
```
oferb: s/proto binary file/binary file

common/FileUtils.java
```diff
@@ -128,4 +130,34 @@ public static String readFileUnchecked(String path) {
...
+  public static void writeProtoBinary(Message proto, String path) throws IOException {
+    mkdirs(path);
+    proto.writeTo(new FileOutputStream(path));
+  }
+
+  /** Writes a proto to proto binary file, rethrows exceptions as unchecked. */
```
oferb: s/proto binary file/binary file

PR #10, opened 2018-04-05
Text differencer for the code review project

yanivsn:
A simple utility for comparing two strings.


common/DifferenceType.java
```diff
@@ -0,0 +1,8 @@
+package com.google.startup.common;
+
+/** Describes the type of the differnce. */
```
oferb: s/differnce/difference

common/CharDifference.java
```diff
@@ -0,0 +1,60 @@
+package com.google.startup.common;
```
oferb: Let's make this a proto

common/DifferenceType.java
```diff
@@ -0,0 +1,8 @@
+package com.google.startup.common;
```
oferb: Let's put this in the proto

common/TextDifferencer.java
```diff
@@ -0,0 +1,104 @@
...
+   *
+   * @param first The first string.
+   * @param second The second string.
+   * @return A list which holds all the text differences.
+   */
+  public static List<CharDifference> getAllTextDifferences(String first, String second) {
+    return getDifferencesFromLCSMatrix(computeLCSMatrix(first, second), first, second);
+  }
+
+  /** Create an empty matrix based on the given dimentions. */
```
oferb: s/dimentions/dimensions

common/TextDifferencer.java
```diff
@@ -0,0 +1,104 @@
...
+   * Compute all the character differences between two strings based on the precomputed Longest
+   * Common Subsequence matrix.
+   *
+   * @param lcsMatrix The precomputed longest subsequence matrix.
+   * @param first The first string of two strings to be compared with.
+   * @param second The second string of two strings to be compared with.
+   * @return A list of all the character differences
+   */
+  private static List<CharDifference> getDifferencesFromLCSMatrix(
+      int[][] lcsMatrix, String first, String second) {
```
oferb: I think it would be more readable to pass arrays from `computeLCSMatrix()`, e.g: `char[] first` (`first.toCharArray()`)

yanivsn:
Updated. PLAT.

common/DifferenceType.java
yanivsn: Done.

common/CharDifference.java
yanivsn: Done.

common/DifferenceType.java
yanivsn: Done.

common/TextDifferencer.java
yanivsn: Done.

common/TextDifferencer.java
yanivsn: @oferb I think it is much more personal preference, but I don't have strong feelings about it, so Done.

common/TextDifferencer.java
```diff
@@ -0,0 +1,123 @@
...
+            second.toCharArray())
+        .stream()
+        .map((builder) -> builder.build())
+        .collect(Collectors.toList());
+  }
+
+  /** Create an empty matrix based on the given dimensions. */
+  private static int[][] createEmptyLCSMatrix(int rowSize, int colSize) {
+    final int[][] lcsMatrix = new int[rowSize][];
+    for (int i = 0; i < lcsMatrix.length; ++i) {
```
oferb: Let's use i++
Also below..

There's no difference in this context, and it's more standard.

common/char_difference.proto
```diff
@@ -0,0 +1,35 @@
...
+/** Describes the type of the difference. */
+enum DifferenceType {
+  // No change has occurred. The source is equal to the
+  // destination.
+  NO_CHANGE = 0;
+
+  // An addition of new characters has occurred.
+  ADDITION = 1;
+
+  // A deletion of characters has occurred. The source is missing several
```
oferb: If CharDifference.difference is only 1 character, why is this plural? "characters", "missing several" (also in ADDITION).

yanivsn:
Done.

PR #12, opened 2018-04-10
Flags cant be empty

vmax:


common/flags/tests/FlagDescTestClass.java
```diff
@@ -18,34 +18,36 @@
 
 import com.google.startupos.common.flags.Flag;
 import com.google.startupos.common.flags.FlagDesc;
-import com.google.common.annotations.VisibleForTesting;
 
 public class FlagDescTestClass {
   @SuppressWarnings("unused")
-  @FlagDesc(name = "string_flag", description = "A flag description")
+  @FlagDesc(name = "string_flag", description = "A flag description", required = false)
```
oferb: Can false be a default value, and so doesn't need to be explicitly specified?

common/flags/tests/FlagsTest.java
```diff
@@ -201,5 +201,32 @@ public void testDoubleFlag() throws Exception {
...
+
+  @Test
+  public void testRequiredStringFlagWithValue() throws Exception {
+    List<String> leftOverArgs =
+        Arrays.asList(
+            Flags.parse(
+                new String[] {"--required_flag", "required"}, Arrays.asList(TESTFLAGS_PACKAGE)));
+
+    assertEquals(0, leftOverArgs.size());
+    assertEquals("flag should have value", "required", FlagDescTestClass.requiredFlag.get());
```
oferb: s/flag/Flag

PR #18, opened 2018-04-16
Improve text difference space and runtime complexity.

yanivsn:
Most code changes require a small change, usually in the middle of the code.
This change suggest truncating the beginning and the end of each string, in case they are equivalent in both given strings. This will reduce the O(MN) calculation only to the necessary part of the diff, causing a significant reduce in both time and space complexity.


common/TextDifferencer.java
```diff
@@ -19,30 +19,108 @@
...
+   * @param offset The offset required for correction.
+   * @return A stream with the corrected offset.
+   */
+  private static Stream<CharDifference.Builder> addOffestTo(
+      final Stream<CharDifference.Builder> stream, final int offset) {
+    return stream.map(diff -> diff.setIndex(offset + diff.getIndex()));
+  }
+
+  /** Return a stream of the non equal char differences of the given two strings. */
+  private static Stream<CharDifference.Builder> getNonMatchingCharDifferences(
```
oferb: I think that if you can make `getDifferencesFromLCSMatrix` work with `int begin, int length`, then you won't need `getNonMatchingCharDifferences` and `addOffestTo`.

Then, if you return a list from `getMatchingCharDifferences`, you can directly add to a list using `addAll` and won't need `mergeDifferences`.

PR #19, opened 2018-04-16
Pre-commit script checking license header and package names

vmax:


pre-commit.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
```
oferb: Can you add documentation on usage? How should this be run?

PR #23, opened 2018-04-19
Add flags parse methods

oferb:


common/flags/Flags.java
```diff
@@ -39,6 +42,27 @@
...
+  }
+
+  /**
+   * Initializes flag values from command-line style arguments.
+   *
+   * @param args command-line arguments to parse values from
+   * @param package1 package root to scan for flags
+   * @param package2 another package root to scan for flags
+   */
+  public static String[] parse(String[] args, Package package1, Package package2) {
```
vmax: @oferb why only 2 but not `Package … packages`?

PR #34, opened 2018-04-28
Add side by side difference to text differencer.

yanivsn:


common/text_difference.proto
```diff
@@ -0,0 +1,46 @@
...
+option java_multiple_files = true;
+
+// Holds information about a text change.
+message TextChange {
+  /** Describes the type of the difference. */
+  enum Type {
+    // Unused.
+    TYPE_UNUSED = 0;
+
+    // No change has occurred. The source is equal to the
```
oferb: Can you take a look at the comments in the deleted file `common/text_differencer.proto` and see if something is relevant there? I tried to simplify and shorten the comments.

PR #41, opened 2018-05-02
Fix `aa init`

vmax:


tools/aa/commands/InitCommand.java
```diff
@@ -33,23 +36,39 @@
...
 
     try {
+      String remoteRepoPath = remoteRepoUrl.get();
       Config config =
           Config.newBuilder()
               .setBasePath(basePath.get())
-              .setRemoteRepoUrl(remoteRepoUrl.get())
+              .addRemoteRepo(
+                  RemoteRepo.newBuilder()
+                      .setId(getNameFromRemoteUrl(remoteRepoPath))
```
oferb: Good idea to get that automatically

tools/aa/commands/InitCommand.java
```diff
@@ -33,23 +36,39 @@
...
+    String path = new URI(remoteRepoUrl.get()).getPath();
+    return path.substring(path.lastIndexOf("/") + 1).replace(".git", "");
+  }
+
   @Override
   public void run(String[] args) {
     Flags.parse(args, InitCommand.class.getPackage());
 
     try {
+      String remoteRepoPath = remoteRepoUrl.get();
```
oferb: I would switch - 
`String remoteRepoId = getNameFromRemoteUrl(remoteRepoUrl.get());`
Then,
```
RemoteRepo.newBuilder()
    .setId(remoteRepoId)
    .setUrl(remoteRepoUrl.get()))
```

tools/aa/AaTool.java
```diff
@@ -17,20 +17,17 @@
...
-import java.text.ParseException;
-import java.util.Arrays;
+import com.google.startupos.tools.aa.commands.CommandsComponent;
+import com.google.startupos.tools.aa.commands.DaggerCommandsComponent;
 import java.util.HashMap;
 
 /** aa tool. */
 public class AaTool {
   private static HashMap<String, AaCommand> commands = new HashMap<>();
+  private static CommandsComponent commandsComponent = DaggerCommandsComponent.create();
```
oferb: Would be cleaner to define one component for the AaTool, here as a nested interface, similar to how it's done at `LocalServer`:
https://github.com/google/startup-os/blob/master/tools/local_server/LocalServer.java#L95-L99

Then, `AaTool` would get `InitCommand` in its constructor.

tools/aa/BUILD
```diff
@@ -19,11 +19,13 @@ java_binary(
     main_class = "com.google.startupos.tools.aa.AaTool",
     deps = [
         ":config_java_proto",
+        "//common",
+        "//common:dagger_with_annotation_processor",
+        "//common/flags",
```
oferb: We probably don't need `"//common/flags",` in `AaTool` anymore..

vmax: @oferb `InitCommand` does use it and it's linked to `aa_tool` as well

oferb: Ohh right..

PR #43, opened 2018-05-02
implement aa workspace

vmax:


common/FileUtils.java
```diff
@@ -182,4 +182,8 @@ public Message readProtoBinaryUnchecked(String path, Message.Builder builder) {
       throw new RuntimeException(e);
     }
   }
-}
\ No newline at end of file
+
+  public void copyDirectoryToDirectory(String source, String destination) throws IOException {
+    org.apache.commons.io.FileUtils.copyDirectory(new File(source), new File(destination));
```
oferb: Valerii is working on making everything here use `fileSystem` so that we can use `jimfs` in tests.
If you don't start from `fileSystem`, you end up using the default filesystem which is not the injected one.

You can look at `folderExists()` as an example of how to use it.

tools/aa/aa_tool.sh
```diff
@@ -0,0 +1,14 @@
...
+function aa {
+    if [ "$1" = "workspace" ]; then
+        AA_RESULT=$(eval $AA_BINARY $*)
+        $AA_RESULT
+    else
+        eval $AA_BINARY $*
+    fi
+}
+
+export -f aa
```
oferb: Newline

tools/aa/aa_tool.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+AA_BINARY="java -jar $AA_BASE/bazel-bin/tools/aa/aa_tool_deploy.jar"
+
+function aa {
+    if [ "$1" = "workspace" ]; then
+        AA_RESULT=$(eval $AA_BINARY $*)
```
oferb: Can you document what's going on here?

tools/aa/commands/ConfigProvider.java
```diff
@@ -0,0 +1,23 @@
+package com.google.startupos.tools.aa.commands;
```
oferb: Let's make this a Module, and then you can inject Config directly.

tools/aa/commands/ConfigProvider.java
```diff
@@ -0,0 +1,23 @@
...
+
+import com.google.startupos.common.FileUtils;
+import com.google.startupos.tools.aa.Protos.Config;
+
+import javax.inject.Inject;
+
+public class ConfigProvider {
+    private FileUtils fileUtils;
+    private static String CONFIG_FILENAME = "~/aaconfig.prototxt";
```
oferb: Let's make this a flag, in the module.

tools/aa/commands/WorkspaceCommand.java
```diff
@@ -0,0 +1,55 @@
...
+import com.google.startupos.common.flags.FlagDesc;
+import com.google.startupos.common.flags.Flags;
+import java.io.IOException;
+import java.nio.file.Paths;
+import javax.inject.Inject;
+
+public class WorkspaceCommand implements AaCommand {
+  @FlagDesc(
+    name = "force",
+    description = "Force workspace switching creating workspace if non-existent"
```
oferb: We wanted to support "-f" as well, right?
Our Flags library doesn't support it (and intentionally, since single-letter flags are more prone to be defined in multiple packages).
To support single-letter flags here, I suggest replacing "-f" with "--force" before passing on to `Flags.Parse()`.

oferb: s/Force workspace switching creating workspace if non-existent/Create workspace if it doesn't exist

tools/aa/commands/WorkspaceCommand.java
```diff
@@ -0,0 +1,55 @@
...
+  )
+  public static Flag<Boolean> force = Flag.create(false);
+
+  @FlagDesc(name = "ws", description = "Workspace name to switch to", required = true)
```
oferb: We'd like to have:
`aa workspace <workspace name>`
`aa workspace -f <workspace name>`

tools/aa/commands/WorkspaceCommand.java
```diff
@@ -0,0 +1,55 @@
...
+    this.fileUtils = utils;
+    this.config = configProvider;
+  }
+
+  @Override
+  public void run(String[] args) {
+    Flags.parse(args, WorkspaceCommand.class.getPackage());
+    String basePath = config.getConfig().getBasePath();
+
+    String user = System.getenv("USER");
```
oferb: We're going to get the user from LocalServer, who gets it from Angular as the current logged-in user.
Until we do that, let's add it to the config and take it from there, but add a TODO to get it from LocalServer.

tools/aa/commands/WorkspaceCommand.java
oferb: s/<current description>/Name of workspace to switch to

tools/aa/commands/ConfigProvider.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.google.startupos.common.CommonModule;
+import com.google.startupos.common.FileUtils;
+import com.google.startupos.tools.aa.Protos.Config;
+import dagger.Module;
+import dagger.Provides;
+import javax.inject.Inject;
+
+@Module(includes = CommonModule.class)
+public class ConfigProvider {
```
elinabtl: It's `ConfigModule`, no?

tools/aa/commands/ConfigProvider.java
```diff
@@ -0,0 +1,37 @@
...
+  @Inject
+  public ConfigProvider() {}
```
elinabtl: Try s/public ConfigProvider/ConfigProvider, see if it still works.

tools/aa/commands/ConfigProvider.java
```diff
@@ -0,0 +1,37 @@
...
+
+  @Provides
+  public static Config getConfig(FileUtils fileUtils, String configFileName) {
+    return (Config)
+        fileUtils.readPrototxtUnchecked(
+            fileUtils.expandHomeDirectory(configFileName), Config.newBuilder());
```
elinabtl: Let's add the `expandHomeDirectory()` inside `readPrototxtUnchecked()`

tools/aa/commands/InitCommand.java
```diff
@@ -36,6 +36,9 @@
   @FlagDesc(name = "remote_repo_url", description = "Remote git repo", required = true)
   public static Flag<String> remoteRepoUrl = Flag.create("");
 
+  @FlagDesc(name = "user", description = "User")
+  public static Flag<String> user = Flag.create(System.getenv("USER"));
```
elinabtl: Nice default :)

tools/aa/commands/WorkspaceCommand.java
```diff
@@ -0,0 +1,81 @@
...
+
+  @Inject
+  public WorkspaceCommand(FileUtils utils, Config config) {
+    this.fileUtils = utils;
+    this.config = config;
+  }
+
+  @Override
+  public void run(String[] args) {
+    for (int i = 0; i < args.length; i++) {
```
elinabtl: Please add a comment for this loop.

tools/aa/commands/WorkspaceCommand.java
```diff
@@ -0,0 +1,81 @@
...
+    if (force.get()) {
+      fileUtils.mkdirs(newWsPath);
+      try {
+        fileUtils.copyDirectoryToDirectory(headPath, newWsPath);
+      } catch (IOException e) {
+        e.printStackTrace();
+      }
+    }
+
+    System.out.println(String.format("cd %s", newWsPath));
```
elinabtl: Please document that this println is going to be executed.
I'd explain it at the top, and then also mention it here.

tools/aa/config.proto
```diff
@@ -17,4 +18,6 @@ message Config {
   string base_path = 1;
   // Remote git repo url
   repeated RemoteRepo remote_repo = 2;
+  // user
```
elinabtl: Please add a TODO to use user from LocalServer.

tools/aa/commands/ConfigProvider.java
vmax: @elinabtl I've checked the source, it in fact is already calling it, see impl [here](https://github.com/google/startup-os/blob/master/common/FileUtils.java#L60)

PR #49, opened 2018-05-08
Add some methods to FileUtils

oferb:


common/FileUtils.java
```diff
@@ -118,6 +118,17 @@ public boolean fileOrFolderExists(String path) {
...
 
+  /** Checks if folder exists. Returns false for files. */
+  public boolean folderEmptyOrNotExists(String path) {
+    if (fileExists(path)) {
+      return false;
+    }
+    if (!folderExists(path)) {
+      return true;
+    }
+    return getFiles(path).isEmpty();
```
val-fed: s/getFiles(path)/listContents(path)

The method has already been renamed. And it throws IOException.

common/FileUtils.java
```diff
@@ -118,6 +118,17 @@ public boolean fileOrFolderExists(String path) {
             || Files.isDirectory(fileSystem.getPath(expandHomeDirectory(path)));
   }
 
+  /** Checks if folder exists. Returns false for files. */
+  public boolean folderEmptyOrNotExists(String path) {
```
val-fed: Add throws IOException

common/FileUtils.java
```diff
@@ -118,6 +118,17 @@ public boolean fileOrFolderExists(String path) {
             || Files.isDirectory(fileSystem.getPath(expandHomeDirectory(path)));
   }
 
+  /** Checks if folder exists. Returns false for files. */
```
val-fed: s/folder/the folder

PR #50, opened 2018-05-08
Add dagger support for TextDifferencer.

yanivsn:
The main changes are in the following order.
- Refactored TextDifferencer into a class.
- Added dagger support for text differencer.


common/AbstractTextDifferencer.java
```diff
@@ -0,0 +1,110 @@
...
+package com.google.startupos.common;
+
+import com.google.startupos.common.TextChange.Type;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import javax.swing.text.Segment;
+
+/** An abstract implementation of the {@link TextDifferencer}. */
```
oferb: As discussed, we need one diff implementation, so we should choose the best one.
WDYT?

PR #52, opened 2018-05-09
Make base folder relative to working folder, and tweak init and workspace

oferb:


tools/aa/commands/WorkspaceCommand.java
```diff
@@ -27,8 +27,14 @@
 import javax.inject.Inject;
```
val-fed: Unused import: java.nio.file.Paths

tools/aa/commands/InitCommand.java
```diff
@@ -25,46 +25,68 @@
 import java.net.URI;
 import java.net.URISyntaxException;
```
val-fed: Unused imports:
* com.google.startupos.tools.aa.Protos.RemoteRepo
* java.net.URI
* java.net.URISyntaxException

tools/aa/AaTool.java
```diff
@@ -43,27 +47,13 @@ private void printUsage() {
...
   }
 
   private void run(String[] args) {
     if (args.length > 0) {
       String command = args[0];
       if (commands.containsKey(command)) {
-        commands.get(command).run(args);
+        commands.get(command).get().run(args);
       } else {
         System.out.println("");
```
val-fed: Maybe needs to add some message instead **""**?

tools/aa/commands/InitCommand.java
```diff
@@ -25,46 +25,68 @@
...
   @FlagDesc(name = "base_path", description = "Base path", required = true)
   public static Flag<String> basePath = Flag.create("");
 
-  @FlagDesc(name = "remote_repo_url", description = "Remote git repo", required = true)
-  public static Flag<String> remoteRepoUrl = Flag.create("");
+  @FlagDesc(name = "startupos_repo", description = "StartupOS git repo")
+  public static Flag<String> startuposRepo = Flag.create("git@github.com:google/startup-os.git");
 
   @FlagDesc(name = "user", description = "User")
   public static Flag<String> user = Flag.create(System.getenv("USER"));
```
val-fed: Maybe, do **basePath, startuposRepo, user variables** as final?

tools/aa/AaTool.java
oferb: It's a newline. The message is printed in `printUsage()`.

tools/aa/commands/InitCommand.java
oferb: Technically you're right, they can be final, but I find it kind of confusing that final fields can be changed if they're objects (i.e the reference cannot change, but the pointed object can).
For this reason, I tend to put final only when it means that nothing can change about the field - i.e the type is immutable.

PR #56, opened 2018-05-11
Implement SyncCommand

vmax:


tools/aa/commands/SyncCommand.java
```diff
@@ -0,0 +1,62 @@
...
+          .listContents(headPath)
+          .stream()
+          .map(path -> fileUtils.joinPaths(headPath, path))
+          .filter(path -> fileUtils.folderExists(path))
+          .map(File::new)
+          .forEach(
+              pathAsFile -> {
+                try {
+                  System.out.println("Performing sync: " + pathAsFile.toString());
+                  System.err.println(Git.open(pathAsFile).pull().call().toString());
```
oferb: How about using `GitRepo.pullAll()`?
https://github.com/google/startup-os/blob/master/common/repo/GitRepo.java
You can see an example here:
https://github.com/google/startup-os/blob/master/common/repo/tests/TestTool.java

You'd have to implement `pull`, but you already have here.

tools/aa/commands/BUILD
```diff
@@ -4,15 +4,16 @@ java_library(
...
     deps = [
-        "//tools/aa:config_java_proto",
         "//common",
         "//common:dagger_with_annotation_processor",
         "//common/flags",
+        "//common/repo",
         "//external:jar/com/google/protobuf/protobuf_java",
         "//third_party/maven/com/google/guava",
         "//third_party/maven/javax/inject:javax_inject",
+        "//third_party/maven/org/eclipse/jgit:org_eclipse_jgit",
```
oferb: Do you actually still need `jgit:org_eclipse_jgit`?

PR #61, opened 2018-05-14
Add InitCommentTest, add FileUtils.listContentsRecursively()

oferb:


vmax:
@oferb instead of "Fix-" commits, it's better to do `git commit --amend` which modifies the previous commit (if you've already pushed it, you'll need to do `git push --force`

tools/aa/commands/InitCommand.java
```diff
@@ -77,14 +77,18 @@ public void run(String[] args) {
...
+      if (!startuposRepo.get().isEmpty()) {
+        // Clone StartupOS repo into head:
+        String startupOsPath = fileUtils.joinPaths(basePath.get(), "head", "startup-os");
+        System.out.println("Cloning StartupOS into " + startupOsPath);
+        Git.cloneRepository()
+            .setURI(startuposRepo.get())
+            .setDirectory(Paths.get(startupOsPath).toFile())
+            .call();
+        System.out.println("Completed Cloning");
+      } else {
```
vmax: @oferb FYI this may only work if you supply `--startupos_repo ""` as cmdline arg. That's OK if it was intended but I'm not sure somebody would do it 🙂

oferb: This is primarily because I didn't know how to test that part :-P

PR #63, opened 2018-05-14
Flags: warn if Flags.parse is not called

vmax:


common/flags/Flags.java
```diff
@@ -79,7 +79,13 @@ public static void printUsage() {
 
   @VisibleForTesting
   static String getFlagValue(String name) {
-    if (!instance().getFlag(name).getHasValue()) {
+    FlagData flagData = instance().getFlag(name);
+    if (flagData == null) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Flag data '%s' is null; did you forget to call FlagDatas.parse()?", flagData));
```
oferb: s/FlagDatas/Flags

PR #65, opened 2018-05-15
Current workspace name provider

vmax:


tools/aa/AaModule.java
```diff
@@ -43,10 +43,29 @@
...
+    Path currentFolder = Paths.get(fileUtils.getCurrentWorkingDirectory());
+    Path wsFolder = Paths.get(fileUtils.joinPaths(basePath, "ws"));
+    try {
+      Path currentWsFolder = wsFolder.relativize(currentFolder).normalize();
+      String wsName = currentWsFolder.subpath(0, 1).toString();
+      if (wsName.equals("..")) {
+        throw new RuntimeException("You're inside base folder but not in a workspace");
+      }
+      return wsName;
+    } catch (IllegalArgumentException ex) {
```
oferb: s/IllegalArgumentException/IllegalStateException

vmax: @oferb 
it's not thrown, see [here](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#subpath(int,%20int)) 🙂
 ```
Throws:
IllegalArgumentException - if beginIndex is negative, or greater than or equal to the number of elements. If endIndex is less than or equal to beginIndex, or larger than the number of elements.
```

PR #68, opened 2018-05-15
Additional functionality for aa sync command

vmax:


tools/aa/commands/SyncCommand.java
```diff
@@ -48,11 +57,49 @@ public void run(String[] args) {
...
+                repo.commit(repo.getUncommittedFiles(), "Sync: temporary commit");
+                System.out.println(
+                    String.format("[%s/%s]: switching to master", currentWorkspaceName, repoName));
+                repo.switchBranch("master");
+                System.out.println(
+                    String.format("[%s/%s]: pulling", currentWorkspaceName, repoName));
+                repo.pull();
+                System.out.println(
+                    String.format("[%s/%s]: merging changes", currentWorkspaceName, repoName));
+                repo.merge("temp_branch_for_sync");
```
oferb: What happens if the merge fails? Will we get <<<>>> in the file?

vmax: @oferb this one is a great finding, thanks. With changes in https://github.com/google/startup-os/pull/68/commits/b8bff6d103962b8966ad8541e26f253 we now will

tools/aa/commands/SyncCommand.java
```diff
@@ -48,11 +57,55 @@ public void run(String[] args) {
...
+                System.out.println(
+                    String.format("[%s/%s]: pulling", currentWorkspaceName, repoName));
+                repo.pull();
+                System.out.println(
+                    String.format("[%s/%s]: merging changes", currentWorkspaceName, repoName));
+                boolean mergeResult = repo.merge("temp_branch_for_sync");
+                if (!mergeResult) {
+                  System.out.println(
+                      String.format(
+                          "[%s/%s]: merge did NOT go clear, check files for conflicts",
```
oferb: s/merge did NOT go clear/manual merge required

tools/aa/commands/SyncCommand.java
```diff
@@ -48,11 +57,55 @@ public void run(String[] args) {
...
           .forEach(
               path -> {
-                System.out.println("Performing sync: " + path.toString());
+                System.out.println(String.format("[HEAD]: Performing sync: %s", path));
                 repoFactory.create(path).pull();
               });
     } catch (IOException e) {
       e.printStackTrace();
     }
+    // then, do the sync for all workspaces
```
oferb: I'd like you to consider what happens if we fail at every step (even if the user pressed `ctrl-c`). Could we make it such that `ctrl-c` leaves no side effects?

For example, if we're writing to a file at location A, we could write to a file in /tmp, and then when we're done, move it from /tmp to A. This doesn't eliminate the possibility of an issue with ctrl-c, since it might prevent us from moving, but move is such a fast operation that this is unlikely to happen.

So please think about this - will ctrl-c leave us in a "bad state"?

oferb: Let's add a TODO for this, following our discussion.

PR #69, opened 2018-05-15
Add dagger support for TextDifferencer.

yanivsn:
#50 Includes irrelevant commits. 

In addition, I restored the LCSTextDifferencer into a single implementing class as requested.


common/TextDifferencer.java
```diff
@@ -16,16 +16,9 @@
...
-import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
-import java.util.stream.Collectors;
-import java.util.stream.IntStream;
-import javax.swing.text.Segment;
 
-/** An implementation of text difference based on the Longest Common Subsequence (LCS) problem. */
-public class TextDifferencer {
+public interface TextDifferencer {
```
oferb: If there's only one implementation (`LCSTextDifferencer`) then you don't need an interface, and then you also don't need `provideDefaultTextDifferencer()`.
If in the future we'll have 2 implementations, we can introduce an interface then.

common/LCSTextDifferencer.java
```diff
@@ -0,0 +1,194 @@
...
+import com.google.startupos.common.TextChange.Type;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import javax.swing.text.Segment;
+
+/** An implementation of text difference based on the Longest Common Subsequence (LCS) problem. */
+public class LCSTextDifferencer implements TextDifferencer {
```
oferb: I see you kept the `TextDifferencer` interface.
If we only have one implementation, we don't need an interface. Not having an interface also saves us from needing a provider, to connect the interface to the class.

PR #72, opened 2018-05-18
googletest and abseil

nlewycky:
Add GoogleTest.
Add Abseil and its dependency Google benchmark.
Opt in to --expand_configs_in_place which will be made the default.


.bazelrc
```diff
@@ -0,0 +1,4 @@
+# This will be the default in bazel.
+# https://blog.bazel.build/2018/01/19/config-parsing-order.html
+startup --expand_configs_in_place
```
oferb: The link (`https://blog.bazel.build/2018/01/19/config-parsing-order.html`) says it should no longer be configurable at `0.13`, which is the latest release. Which bazel version are you running?

WORKSPACE
```diff
@@ -111,3 +111,21 @@ http_file(
     executable = True,
     sha256 = "dd8d58429a258b094b20a1435be3086ecee5d036b87c0e766309842766bc345b"
 )
+
+http_archive(
+    name = "com_google_googletest",
+    strip_prefix = "googletest-master",
+    urls = ["https://github.com/google/googletest/archive/master.zip"],
```
oferb: If we work with `master`, the repo is no longer reproducible, since if I run it and it works, when you run it, `head` may change at that repo and it doesn't work.
How about we use a release like `com_google_protobuf` or a commit hash like here?
https://github.com/hasadna/hasadna/blob/master/WORKSPACE#L38-L42

WORKSPACE
```diff
@@ -111,3 +111,21 @@ http_file(
     executable = True,
     sha256 = "dd8d58429a258b094b20a1435be3086ecee5d036b87c0e766309842766bc345b"
 )
+
+http_archive(
```
oferb: I'm wondering - what's the standard equivalent of Maven (package management) in cpp?
If 2 dependencies both depend on dep A, but with different versions, how is that resolved?

I see that Abseil recommend "living at HEAD", which I like, but not everybody does that :)
https://abseil.io/about/philosophy#we-recommend-that-you-choose-to-live-at-head

.bazelrc
nlewycky: Looks like I built it from head shortly pre-0.11. I'll update, thanks!

WORKSPACE
nlewycky: Worse problem: Google Test hasn't had a release since adding BUILD files. (Or, since 2016 for that matter.)

But yes, the live-at-head vs. reproducibility issue is a problem. I think the right fix is to pin the other repository by commit ID, and have an update or sync tool which updates your workspace to the latest commit ID and then you commit that workspace change to your local repository. However, Bazel recommends [new_]http_archive over [new_]git_repository. I don't really understand the motivation behind that or how they expect this to work in the larger ecosystem.

WORKSPACE
nlewycky: There is no common package management for C++. Bazel tries to do its own, the Buck build system has Buckaroo, Visual Studio has vcpkg. There's been a recent boom in C++ package managers in the past year or so, but it's too early for there to be any clear winners or losers yet.

Diamond dependencies are yet another unsolved problem. I know some library projects will go out of their way to rewrite things instead of using libraries, so that programs which use them are less likely to have a diamond dependency problem. Barring that, I would hope the libraries are written in such a way that two copies can coexist (no program-scope globals, no global constructors), barring that I would hope that updating the codebase that depends on the older one to use the newer one isn't that hard. For a much longer discussion, see Titus Winters' talk at cppcon, https://www.youtube.com/watch?v=tISy7EJQPzI&t=15m0s when announcing Abseil.

nlewycky:
Answering questions but not updating patch yet.

WORKSPACE
oferb: I agree - so how about we use the http_archive with commit ID, like in:
https://github.com/hasadna/hasadna/blob/master/WORKSPACE#L38-L42

WORKSPACE
oferb: :-)

.bazelrc
oferb: Sure thing ;)

WORKSPACE
ras0219-msft: >Visual Studio has vcpkg

Dev on vcpkg here! We hope Vcpkg is useful to all C++ developers on all platforms, not just Visual Studio users. We recently expanded to support Linux and OSX and we know others have successfully used us to target more exotic platforms like Android, iOS, and Emscripten.

If you have any feedback for us about something we could do to make vcpkg more useful for you, I'd love to hear it!

oferb: Hey neighbor! Haven't tried it and I don't do much C++ development these days.
Thanks for dropping by!

WORKSPACE
nlewycky: Done.

WORKSPACE
nlewycky: Done!

ras0219-msft, sorry to suggest that vcpkg was tied to Visual Studio! I know it isn't, what I meant to say was that it came from the VC++ team as that team's solution to C++ packaging, not that it's tied to Visual Studio. FWIW, I've heard good things about vcpkg!

.bazelrc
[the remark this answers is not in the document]
nlewycky: Removed.

nlewycky: Removed.

nlewycky:
I signed it!

oferb:
Approved.
cla/google is still failing..
Maybe your email is not set on your commits? See _Individual signers_ section above.

nlewycky:
I signed it!

PR #85, opened 2018-05-24
Add sha256 entries for all http_archive's.

nlewycky:


nlewycky:
Er, I don't have a particularly good technique, I ran wget on the urls to download the files, ran sha256sum over them and pasted the results into the workspace file. Do you want that written down? If so, where?

oferb:
If you can write that down in the WORKSPACE file, that'd be great. Thanks!!

nlewycky:
Done?

oferb:
Yup :)

PR #86, opened 2018-05-24
Add required deps to get flogger running

oferb:
Binary targets were found using:
`bazel query //... --output label_kind | grep java_binary`


dependencies.yaml
```diff
@@ -154,3 +157,8 @@ dependencies:
     reflections:
       lang: java
       version: "0.9.11"
+
+  org.slf4j:
+    slf4j-simple:
+      lang: java
+      version: "1.7.25"
```
jbduncan: @oferb @vmax I did a search for `slf4j` in Flogger (https://github.com/google/flogger/search?q=slf4j&unscoped_q=slf4j), and apparently there are no results, so I'm a bit confused as to why slf4j-simple is required here.

Can either of you clarify things for me?

oferb: Sure.
We had it originally, then removed it when moving to Flogger, and then started getting warnings from libraries that are using it. Following is such an error when running the `diff` command of `aa_tool`.

```
INFO: Analysed target //tools/aa:aa_tool (2 packages loaded).
INFO: Found 1 target...
Target //tools/aa:aa_tool up-to-date:
  bazel-bin/tools/aa/aa_tool.jar
  bazel-bin/tools/aa/aa_tool
INFO: Elapsed time: 3.439s, Critical Path: 0.86s
INFO: 3 processes: 1 linux-sandbox, 1 local, 1 worker.
INFO: Build completed successfully, 7 total actions
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
...
```

oferb: My guess is that one of the deps we bring in from Maven is legitimately using it.

PR #91, opened 2018-05-25
Editable description

MateMahinKhan:
I removed the bug component and used a generic component `EditablePropertyComponent` for both Description and Bug property because they were very similar.

We can create a directive for some of the repetitive stuff.


tools/reviewer/web/src/app/dashboard/index.ts
```diff
@@ -2,33 +2,31 @@ export * from './reviews-panel/';
...
   CodeBlockComponent,
   CommentsComponent,
   DiffComponent,
   DiffService,
+  EditablePropertyComponent,
   LineNumbersComponent,
   PersonListComponent,
   ReviewComponent,
-  ReviewsPanelComponent,
+  ReviewsPanelComponent
```
webdevelopland: I noticed that you don't like the extra comma in lists.
But think about it.
It's not an error and doesn't cause any issues. But this makes it easier to work with lists. Each line has the same structure. When you want to duplicate the last line (Ctrl + Shift + D in sublime) to add new one you don't need to add a comma to the previous line. It's convenient.
Also, it's pretty common practice.
Look, angular developers use it:
https://github.com/alexeagle/angular-bazel-example/blob/master/src/app.module.ts#L11

tools/reviewer/web/src/app/dashboard/reviews-panel/review/editable-property/editable-property.component.spec.ts
```diff
@@ -1,20 +1,20 @@
...
     TestBed.configureTestingModule({
-      declarations: [ BugComponent ]
+      declarations: [ EditablePropertyComponent ]
     })
     .compileComponents();
   }));
 
   beforeEach(() => {
-    fixture = TestBed.createComponent(BugComponent);
+    fixture = TestBed.createComponent(EditablePropertyComponent);
```
webdevelopland: Do you think there's any reason to have unit tests of components?
Thing is, they are usually huge because you have to import all angular interface modules there. You have to import them inside your app module, and inside all your spec files too. It makes code less DRY, that's why it becomes more and more difficult to maintain the unit tests. But what's more important, the tests don't make much sense. All complex logic should be inside services (local and global) and if a component has an error, you already get a message in your console.

Please, let me know your thoughts about it.

MateMahinKhan: Yeah, I haven't seen component tests doing much good other than just loading and saying all is good.

tools/reviewer/web/src/app/dashboard/index.ts
MateMahinKhan: Hey Vadim, I love discussing grammar :). I'm of the opinion that code should always follow grammar like it's speaking a language. The list should be alphabetized so that means most of the times the new entry will be some where in the middle of the list. If you want to add an entry at the end of the list, I always (Ctrl + Shift +D) the second last line in the list. It has the same effect of convenience. 

If you put a comma at the end of a json file it'll start giving error so it's a nice practice to always keep it nice and clean. A nice practice of dotting your i's crossing your t's and semicoloning your JavaScript statements :D

PR #92, opened 2018-05-26
Add aa review

oferb:


tools/aa/commands/DiffCommand.java
```diff
@@ -79,9 +103,9 @@ private Diff createDiff() {
             .setWorkspace(currentWorkspaceName)
             .setDescription(description.get())
             .setBug(buglink.get())
-            .addAllReviewer(Arrays.asList(reviewers.get().split(",")))
+            .addAllReviewer(getReviewers(reviewers.get()))
             .setNumber(response.getLastDiffId())
-            .addAuthor(config.getUser());
+            .setAuthor(Author.newBuilder().setName(config.getUser()).build());
```
vmax: @oferb I think in protos you can set `Builder` instances as fields, so you should be able to omit `.build()` here

oferb: I checked before and it didn't work, should check again. Let's leave this one as is for now (hard for me to check it now).
Good comment though.

PR #95, opened 2018-05-26
Fix FileUtils cwd when run from bazel run; Make all tests green

oferb:


commands.txt
```diff
@@ -1,7 +1,5 @@
-# Build every buildable Bazel target
-# Run this to ensure that everything still builds, kind of a regression test
-# This can take several minutes depending hardware and repo size
-bazel query //... | grep -v ^\/\/third_party | xargs bazel build
+# Build every buildable Bazel target and test every testable target
+bazel query //... | grep -v ^\/\/third_party | xargs bazel test
```
vmax: @oferb does `bazel test` work on binaries?

oferb: It builds all targets, but runs `bazel test` only on test targets

PR #98, opened 2018-05-27
Fix bazel query by including relevant repos in workspace

vmax:
`FIXME`: this one is _temporary_ until we move to Angular Bazel


WORKSPACE
```diff
@@ -154,3 +154,30 @@ http_archive(
     strip_prefix = "abseil-cpp-59ae4d5a0e833bedd9d7cc059ac15a9dc130e3f7",
     urls = ["https://github.com/abseil/abseil-cpp/archive/59ae4d5a0e833bedd9d7cc059ac15a9dc130e3f7.zip"],
 )
+
+
+http_archive(
```
oferb: Can you add a TODO:
// TODO: Once we move to Angular Bazel, remove all below (and changes in commit that added it).

Also a reminder in `commands.txt` to run `npm install @bazel/typescript`.

PR #103, opened 2018-05-27
Add extra action to produce clang compilation database.

nlewycky:
Clang compilation database is documented here: https://clang.llvm.org/docs/JSONCompilationDatabase.html

As provided, this isn't immediately useful. Either Bazel will need to emit the path to the source file in a WORKSPACE-relative form so that we can run tools that operate over the whole codebase at once, or you would have to edit update_compile_database.cc to integrate with the clang tool (such as running "clang-tidy", which would then also need to be available in the sandbox) and it could only operate on one source file at a time.

### DIFF
--- a/third_party/BUILD.jsoncpp
+++ b/third_party/BUILD.jsoncpp
@@ -1,0 +1,110 @@
+# JsonCpp is a C++ library that allows manipulating JSON values, including
+# serialization and deserialization to and from strings.
+
+licenses(["unencumbered"])
+
+exports_files(["LICENSE"])
+
+JSONCPP_VERSION_DEFINES = [
+    "-DJSONCPP_VERSION_STRING=1.8.4",
+    "-DJSONCPP_VERSION_MAJOR=1",
+    "-DJSONCPP_VERSION_MINOR=8",
+    "-DJSONCPP_VERSION_PATCH=4",
+]
+
+cc_library(
+    name = "jsoncpp",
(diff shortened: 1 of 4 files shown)


tools/clang_json/update_compile_database.cc
```diff
@@ -0,0 +1,55 @@
...
+
+#include "google/protobuf/io/coded_stream.h"
+#include "google/protobuf/io/zero_copy_stream_impl.h"
+#include "google/protobuf/stubs/common.h"
+
+#include "src/main/protobuf/extra_actions_base.pb.h"
+
+#include "include/json/json.h"
+
+int main(int argc, char **argv) {
```
oferb: Can you add some documentation for what this does?

third_party/BUILD.jsoncpp
```diff
@@ -0,0 +1,107 @@
+licenses(["unencumbered"])
```
oferb: Can you add documentation for what this does? Either here or in `startupos_external_jsoncpp` in `WORKSPACE`.

tools/clang_json/update_compile_database.cc
nlewycky: Done.

third_party/BUILD.jsoncpp
nlewycky: Done.

PR #104, opened 2018-05-27
Add aa snapshot, change aa diff to just track diff branch

oferb:


common/repo/tests/MemoryRepo.java
```diff
@@ -78,6 +78,10 @@ public void switchBranch(String branch) {
     currentBranch = branch;
   }
 
+  public void tagHead(String name) {
+    throw new UnsupportedOperationException("Not implemented");
```
vmax: @oferb how about making `HashMap<Commit, String> tags` to support it?

tools/aa/aa_tool.sh
```diff
@@ -71,9 +71,9 @@ function aa {
   STARTUP_OS=$AA_BASE/head/startup-os
 
   # Uncomment to override StartupOS repo:
-  #STARTUP_OS=<repo path>
+  STARTUP_OS=/home/oferb/devel/base/ws/aa_checks2/startup-os
```
vmax: @oferb I know it's a draft, but you utilize something like [direnv](https://direnv.net/) to switch envs as you switch directories when testing locally

common/repo/tests/MemoryRepo.java
oferb: I'd rather not invest in MemoryRepo, until we actually start using it.

PR #106, opened 2018-05-30
Add support for lists to flags

val-fed:


common/flags/Flag.java
```diff
@@ -44,6 +48,26 @@
     return new Flag.DoubleFlag(defaultValue);
   }
 
+  public static Flag<List<String>> createStringsListFlag(List<String> defaultValue) {
```
oferb: I think you can name all of them `create()` since they differ in their inputs.

common/flags/Flag.java
```diff
@@ -18,11 +18,15 @@
...
 import com.google.common.flogger.FluentLogger;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
 
 /** Flag class, with implementations for various flag types. */
-// TODO: Add support for lists
 public abstract class Flag<T> {
   private static final FluentLogger log = FluentLogger.forEnclosingClass();
+  private static final String STRING_DELIMITER = ",\\s*";
```
oferb: Was `split(",")` not enough?

common/flags/Flags.java
```diff
@@ -104,7 +104,13 @@ static void setFlagValue(String name, String value) {
 
   @VisibleForTesting
   static String getDefaultFlagValue(String flagName) {
-    return instance().getFlag(flagName).getDefault();
+    FlagData flagData = instance().getFlag(flagName);
+    if(flagData.getIsListFlag()){
```
oferb: Please format as:
`if (flagData.getIsListFlag()) {`
(note the spaces)
Also elsewhere.

common/flags/Flags.java
```diff
@@ -104,7 +104,13 @@ static void setFlagValue(String name, String value) {
...
+      // deleting characters "[" and "]" at the beginning and end of the List
```
oferb: Why would we have `"[" and "]"` at the beginning and end of the list?

common/flags/Flag.java
val-fed: It's impossible to use one method's name for all methods which work with Lists<> because generic types are erased at runtime, so all of these methods compile to the same signature. If we want to use one name we need to think it over how to do it.

common/flags/Flags.java
val-fed: We have these symbols from a default value of the flag of a list. 
When we create a new flag of the list:
```
@FlagDesc(name = "strings_list_flag", description = "A flag description")
  public static final Flag<List<String>> stringsListFlag
      = Flag.createStringsListFlag(Arrays.asList("ab", "cd"));
```
And we want to get a default value we will get `[ab, cd]` because `toString()` method is called. This call happens here - https://github.com/google/startup-os/blob/4aa5dcff4545d276190a8501b58ec8666f3109d8/common/flags/ClassScanner.java#L188

common/flags/Flag.java
```diff
@@ -18,11 +18,15 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
 
 /** Flag class, with implementations for various flag types. */
-// TODO: Add support for lists
 public abstract class Flag<T> {
   private static final FluentLogger log = FluentLogger.forEnclosingClass();
+  private static final String STRING_DELIMITER = ",";
+  private static final String NUMBER_AND_BOOLEAN_DELIMITER = "\\s*,\\s*";
```
oferb: Why do numbers and booleans haver a different delimiter than ","?

common/flags/Flag.java
oferb: Interesting, though are you sure about this? See these:
https://stackoverflow.com/questions/3452859/java-generics-code-compiles-with-javac-fails-with-eclipse-helios

https://stackoverflow.com/questions/5527235/type-erasure-and-overloading-in-java-why-does-this-work?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

PR #107, opened 2018-05-31
Add support for parsing for a specific class in Flags

val-fed:


common/flags/ClassScanner.java
```diff
@@ -81,7 +78,19 @@ private JarFile getJarFile(URL url) throws IOException {
     return urlConnection.getJarFile();
   }
 
-  private ImmutableList<Field> getFields(String packageName) throws IOException {
+  private ImmutableList<Field> getClassFields(Class clazz){
+    ImmutableList.Builder<Field> result = ImmutableList.builder();
+    Field[] fields = clazz.getDeclaredFields();
+    for(Field field: fields){
+      FlagDesc flagDesc =  field.getAnnotation(FlagDesc.class);
```
oferb: Let's check that `field` is of type `Flag`, as in `getPackageFields()`.

common/flags/ClassScanner.java
```diff
@@ -128,8 +137,28 @@ private JarFile getJarFile(URL url) throws IOException {
     return result.build();
   }
 
+  public void scanClass(Class clazz, Map<String,FlagData> flags) {
```
oferb: Since most of the code is the same, it'd be good to try and share code with `scanPackage()` instead of duplicate.

common/flags/tests/FlagsTest.java
```diff
@@ -232,7 +232,199 @@ public void testRequiredStringFlagWithValue() throws Exception {
   }
 
   @Test(expected = IllegalArgumentException.class)
-  public void testFlagThrowsWithoutParsing() throws Exception {
+  public void testFlagThrowsWithoutParsing() {
     FlagDescTestClass.integerFlag.get();
   }
+
+  @Test
```
oferb: No need to duplicate all tests, since most of the functionality is already tested.
Since you added code that takes flags from a class, here are some things I'd test:
* Find flag in class
* Not fail when there is no flag in class
* Not take flag in package but in another class

PR #112, opened 2018-06-06
CircleCI integration

vmax:


.circleci/config.yml
```diff
@@ -0,0 +1,77 @@
...
+
+# These jobs will run in parallel, and report separate statuses to GitHub PRs
+jobs:
+  lint:
+    <<: *job_defaults
+    steps:
+      - checkout:
+          <<: *post_checkout
+      - run: './pre-commit.sh'
+      - run: 'buildifier -mode=check $(find . -type f \( -name BUILD.bazel -or -name BUILD \)) ||
```
oferb: Does `yarn buildifier` run the same buildifier as this `buildifier` in the docker image?

vmax-reef: @oferb not really, `buildifier` comes bundled with docker image but in `yarn buildifier` we download it. so it may be a good idea to replace CI command with `bazel run //tools:buildifier`

vmax: @oferb not really, `buildifier` comes bundled with docker image but in `yarn buildifier` we download it. so it may be a good idea to replace CI command with `bazel run //tools:buildifier`

PR #113, opened 2018-06-06
Fix bugs with getting item of list flag and logging

val-fed:


common/flags/Flags.java
```diff
@@ -100,7 +100,15 @@ static String getFlagValue(String name) {
     if (!flagData.getHasValue()) {
       return null;
     }
-    return instance().getFlag(name).getValue();
+    if (flagData.getIsListFlag()){
+      return flagData.getValue()
+          // deleting characters "[" and "]" at the beginning and end of the List and spaces after commas
+          // for correct parse values
+          .replace("[", "")
```
oferb: Didn't you remove this in the previous PR in `getDefaultValue`?

val-fed: Yes, I first deleted it. But after we decided not to support the commas at `strings list flag`, I decided that it would be better to do this in one place than to override `getDefault()` method for each list flag.

oferb: I think it's better to store it correctly in the first place, than to remember to correct it every time the data is read - or maybe there's some issue with correcting it at `getDefault()`?

common/flags/Flags.java
```diff
@@ -110,21 +110,17 @@ static FlagData getFlag(String name) {
 
   @VisibleForTesting
   static void setFlagValue(String name, String value) {
-    instance().flags.put(name, getFlag(name).toBuilder().setValue(value).setHasValue(true).build());
+    instance().flags.put(name, getFlag(name).toBuilder()
+        .setValue(value
+            .replace("[", "")
```
oferb: Looks like we're removing `[` and `]` and spaces before commas from all flags. What if I want to pass a string flag that has `[`?

PR #116, opened 2018-06-07
Add example k8s services

lewish:
Providers a container_push bazel rule for pushing a container to a container registry, and an example kubernetes config file defining a deployment, and an exposed service running the hello world example.


lewish:
I signed it!

PR #117, opened 2018-06-07
Add methods for removing directories and files to FileUtils

val-fed:


common/FileUtils.java
```diff
@@ -269,4 +269,49 @@ public FileVisitResult visitFile(final Path file, final BasicFileAttributes attr
...
+  public void deleteDirectoryContentUnchecked(String path){
+    try {
+      deleteDirectoryContent(path);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  /** Deletes file or folder. */
+  public void deleteFileOrDirectoryIfExist(String path) throws IOException {
```
oferb: s/deleteFileOrDirectoryIfExist/deleteFileOrDirectoryIfExists

common/FileUtils.java
```diff
@@ -269,4 +269,49 @@ public FileVisitResult visitFile(final Path file, final BasicFileAttributes attr
   public void copyDirectoryToDirectory(String source, String destination) throws IOException {
     copyDirectoryToDirectory(source, destination, null);
   }
+
+  /** Deletes all files and folders in directory. */
```
oferb: Does it delete the directory itself?

val-fed: Yes, target directory is deleted also.

PR #119, opened 2018-06-10
Implement getCommits() in GitRepo

oferb:


common/repo/GitRepo.java
```diff
@@ -21,17 +21,24 @@
...
+import com.google.common.collect.Lists;
 import com.google.startupos.common.repo.Protos.Commit;
 import com.google.startupos.common.repo.Protos.File;
 import java.io.IOException;
 import java.nio.file.FileSystem;
 import java.nio.file.Paths;
+import java.util.Date;
+import java.util.List;
+
+import jdk.nashorn.internal.ir.annotations.Immutable;
```
jbduncan: Oops, I think the wrong `@Immutable` annotation was imported here. Did you by any chance mean to import error-prone's annotation rather than the JDK's internal one for Nashorn? :)

PR #121, opened 2018-06-10
WorkspaceCommand: ability to remove workspaces

vmax:


tools/aa/commands/WorkspaceCommand.java
```diff
@@ -103,6 +113,10 @@ public boolean run(String[] args) {
     String workspacePath = fileUtils.joinPaths(basePath, "ws", workspaceName);
 
     if (force.get()) {
+      if (remove.get()) {
```
oferb: Let's put this in `checkFlags()`

tools/aa/commands/WorkspaceCommand.java
```diff
@@ -121,6 +135,16 @@ public boolean run(String[] args) {
         System.err.println(RED_ERROR + "Workspace does not exist");
         return false;
       }
+
+      if (remove.get()) {
+        Path currentPath = Paths.get("").toAbsolutePath();
+        if (currentPath.startsWith(workspacePath)) {
+          System.err.println(RED_ERROR + "Trying to remove active workspace");
```
oferb: Why not allow removing active workspace?
In bash, you can delete the current folder.

vmax: @oferb but is this a feature and not a bug? 🙂 
apart from `cd`-ing away, you won't be able to do much being left in this state

tools/aa/commands/WorkspaceCommand.java
```diff
@@ -42,93 +43,162 @@
...
     }
 
     args = Arrays.copyOfRange(args, 0, args.length - 1);
-
     Flags.parse(args, WorkspaceCommand.class.getPackage());
-    String basePath = config.getBasePath();
-    String workspacePath = fileUtils.joinPaths(basePath, "ws", workspaceName);
+
+    if (force.get() && remove.get()) {
+      System.err.println(RED_ERROR + "Conflicting options specified");
```
oferb: How about:
`Cannot specify both force and remove flags`

tools/aa/commands/WorkspaceCommand.java
```diff
@@ -42,93 +43,162 @@
...
+
+  @Override
+  public boolean run(String[] args) {
+    // Note: System.out gets executed by the calling aa_tool.sh to run commands such as cd.
+    if (!processArgs(args)) {
+      return false;
+    }
+
+    if (mode == ActionMode.CREATE) {
+      if (!createWorkspace()) {
```
oferb: How about we put the positive first:
```
if (createWorkspace()) {
  switchWorkspace();
} else {
  return false;
}
```
(also below)

vmax: yay, we saved space on two `!` characters 🙂

oferb: :-)

PR #122, opened 2018-06-11
Fix imports in GitRepo

oferb:


jbduncan:
LGTM! :+1:

oferb:
Done in https://github.com/google/startup-os/pull/123 instead.

PR #126, opened 2018-06-12
Reviewer - protoc

webdevelopland:
Main changes:
- `google-protobuf compiler` and `ts-protoc-gen` are used instead of `protobuf.js`
- code formatting


vmax:
@googlebot rescan

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/code-block/code-block.component.ts
```diff
@@ -37,7 +37,17 @@ export class CodeBlockComponent implements OnInit, OnDestroy {
   @Input() fileContent: string;
   @Input() isNewCode: boolean;
   @Input() changes: number[];
-  // TODO: Do not use Threads
+  // TODO: add supporting of deleting whole thread.
```
oferb: Why do we have all these issues? Is it because we're using protobins somehow?

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/code-block/comments/comments.component.ts
```diff
@@ -89,11 +89,11 @@ export class CommentsComponent implements OnInit {
   }
 
   addComment(): void {
-    const comment: Comment = {
-      content: this.textareaControl.value,
-      createdBy: this.authService.user,
-      timestamp: Date.now()
-    };
+    const comment = new Comment();
```
oferb: There's no `Common.newBuilder()` pattern? In Java you do:
```
const comment = Comment.newBuilder()
    .setContent(...)
    .setCreatedBy(...)
    .build();
```

tools/reviewer/web/src/app/dashboard/reviews-panel/review/person-list/person-list.component.ts
```diff
@@ -25,24 +25,58 @@ export class PersonListComponent {
...
     // Trim white spaces
-    this.diff[this.property] = persons
-      .map(v => v.trim())
-      .filter(v => v.length);
-    if (this.diff[this.property].length === 0) {
+
+    // WARNING: Bad practice
+    // Do not use: object[propertyName]
+    // - It forces to use any.
+    // - It causes undeceted errors.
```
oferb: undetected, difficult

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review.component.html
```diff
@@ -1,31 +1,34 @@
-<div *ngIf="diff" class="single-review">
+<!-- This html file is too big -->
+<!-- TODO: do something with that -->
```
oferb: Let's combine the 2 lines:
`!-- TODO: Make this html file smaller, or split into separate files -->`

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review.component.ts
```diff
@@ -67,26 +62,31 @@ export class ReviewComponent implements OnInit {
...
   removeProperty(property: string, message: string): void {
-    this.firebaseService
-      .removeProperty(this.diff, property)
-      .then(() => {
-        this.notificationService.success(message);
-      })
-      .catch(() => {
-        this.notificationService.error('Some error occured');
-      });
+    console.warn('Unsupported behavior');
```
oferb: What functionality is now broken?

tools/reviewer/web/src/app/dashboard/reviews-panel/reviews-panel.component.ts
```diff
@@ -31,6 +28,10 @@ export class ReviewsPanelComponent implements OnInit {
         if (user) {
           const email = user.email;
           this.username = email.split('@')[0];
+          // TODO: Maybe to use email instead of username?
```
oferb: Why would it be better?

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/code-block/code-block.component.ts
webdevelopland: Actually no.
The issues were present in the last update already. I thought to fix them by adding a feature to update a specific comment instead of a whole diff.
But now, when I realized that we going to keep proto binary in firebase, I added more info about the issues.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review.component.ts
webdevelopland: When proto binary will be used, we won't able to edit a specific property.
So basically we need to write a new method which edits local diff and sends it to the server. I left TODO comment about it.

tools/reviewer/web/src/app/dashboard/reviews-panel/reviews-panel.component.ts
webdevelopland: I based my opinion on an idea, that there could be different domains. E.g.:
oferb@gmail.com
oferb@fakemail.com

But now I realized, that it's always gmail.com. (and maybe google.com..)
So yeah, username is shorter.

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/code-block/code-block.component.ts
oferb: Why doesn't changing the diff locally (i.e adding/removing comments and threads) and updating the whole diff to Firestore work?

webdevelopland: It works! Actually, it's exactly how it implemented here.
We just need to add one new feature. At the moment the app just doesn't know what it should to do, when it gets a message from firebase, that whole thread was deleted.

Two other things are not issues. It's a notice for developers who will be implement deleting of comments.

oferb: Got it, thanks :)

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review.component.html
MateMahinKhan: I changed a lot in this file, actually most of it has been changed. It got merged a couple days ago.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review.component.ts
MateMahinKhan: Sorry If I missed a discussion but we came to the opinion that updating the whole diff is too broad, I think that's why we used to remove specific property?

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review.component.html
webdevelopland: @MateMahinKhan what do you mean? The diff based on fresh state of master.
There's still the huge html file:
https://github.com/google/startup-os/blob/master/tools/reviewer/web/src/app/dashboard/reviews-panel/review/review.component.html

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review.component.ts
webdevelopland: Right. I was talking about the idea in my last diff. It'd be simpler to delete a specific comment, instead of updating whole diff.
But since we're going to use proto binary, that doesn't make a sense. We have to update a whole diff.

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/code-block/code-block.component.ts
webdevelopland: @MateMahinKhan deleting a thread doesn't break proto structure. We aren't deleting all threads, we're just deleting an element.

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/code-block/comments/comments.component.ts
webdevelopland: I think there's no ability like that :(
```typescript
setContent(value: string): void;
```

set functions return nothing.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/person-list/person-list.component.ts
webdevelopland: @MateMahinKhan It's really difficult to maintain big files, components, chunks of code, etc...

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/code-block/code-block.component.ts
MateMahinKhan: Okay. so using proto binary will change this anyway.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review.component.html
MateMahinKhan: @webdevelopland  hmm. so that means I still have these changes. merging will be a tough one for me :D

tools/reviewer/web/src/app/dashboard/reviews-panel/review/person-list/person-list.component.ts
MateMahinKhan: @webdevelopland  here there were be only 2-3 differences and using `object[propertyName]` or a switch case like you used enables to use same component for two properties. It really comes down to developer preference.

PR #128, opened 2018-06-13
PatchCommand: implementation

vmax:


tools/aa/commands/PatchCommand.java
```diff
@@ -67,7 +67,15 @@ public boolean run(String[] args) {
           .forEach(
               path -> {
                 GitRepo repo = this.gitRepoFactory.create(path);
-                repo.merge(branchName, true);
+                String currentBranchName = repo.currentBranch().trim();
```
oferb: Let's do the trim inside `currentBranch()`

PR #130, opened 2018-06-14
Switch some GitRepo commands to call git instead of jGit

oferb:


common/repo/tests/GitRepoTest.java
```diff
@@ -0,0 +1,75 @@
+package com.google.startupos.common.repo.tests;
+
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
```
val-fed: Unused imports:
- import static org.junit.Assert.assertArrayEquals;
- import static org.junit.Assert.assertTrue;
- import java.nio.file.Paths;

common/repo/GitRepo.java
```diff
@@ -48,14 +51,25 @@
...
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+
 
 // TODO: Implement methods
 @AutoFactory
 public class GitRepo implements Repo {
   private Git jGit;
   private Repository jGitRepo;
+  private String repoPath;
```
val-fed: `repoPath` variable is used only in the constructor. Maybe can it be removed?

PR #131, opened 2018-06-15
Change more GitRepo methods to use git instead of jGit

oferb:


common/repo/GitRepo.java
```diff
@@ -118,18 +114,29 @@ private CommandResult runCommand(String... command) {
...
-    return String.format(
+    StringBuilder result = new StringBuilder();
+    result.append(String.format(
         "\n%s\n%s",
         commandResult.command,
-        commandResult.stderr);
+        commandResult.stderr));
+    if (commandLog.size() > 0) {
+      result.append("Previous git commands (most recent is on top):\n");
+      for (CommandResult previousCommand : Lists.reverse(commandLog)) {
```
val-fed: `previousCommand` variable is never used.

oferb: Thanks!

PR #134, opened 2018-06-18
Make code in Common cleaner

val-fed:


common/TextDifferencer.java
```diff
@@ -111,6 +111,7 @@ private static int getFooterMatchingCharactersLength(
    * Generate matching text changes for the given range. The implementation assumes that all the
    * characters within the given range are equal.
    *
+   * // TODO Check params description
```
oferb: Can you elaborate? What kind of check?

common/flags/Flags.java
```diff
@@ -91,10 +90,11 @@ public static void printUsage() {
 
   @VisibleForTesting
   static String getFlagValue(String name) {
-    FlagData flagData = instance().getFlag(name);
+    FlagData flagData = getFlag(name);
     if (flagData == null) {
       throw new IllegalArgumentException(
           String.format(
+              // TODO flagData is always null. Think of a more informative message
               "Flag data '%s' is null; did you forget to call FlagData.parse()?", flagData));
```
oferb: s/flagData/name (in `String.format()`).
Also, let's also mention forgetting to add the package:
Let's change:
did you forget to call FlagData.parse()?
To:
did you forget to call FlagData.parse() or add the package for the flag?

common/flags/Flags.java
```diff
@@ -119,26 +119,26 @@ static void setFlagValue(String name, String value) {
...
   static String getDefaultFlagValue(String flagName) {
-    return instance().getFlag(flagName).getDefault();
+    return getFlag(flagName).getDefault();
   }
 
   private void scanPackages(Iterable<String> packages) {
     for (String packageName : packages) {
       try {
-        classScanner.scanPackage(packageName, flags);
+        classScanner.scanPackage(packageName, getFlags());
```
oferb: Why change from `flags` to `getFlags()`?

common/flags/tests/FlagsTest.java
```diff
@@ -49,7 +50,7 @@ public void defaultsTest() {
     assertEquals(FLAG_SHOULD_HAVE_VALUE, new Integer(123), FlagDescTestClass.integerFlag.get());
     assertEquals(FLAG_SHOULD_HAVE_VALUE, new Long(123456789L), FlagDescTestClass.longFlag.get());
     assertEquals(FLAG_SHOULD_HAVE_VALUE, new Double(1.23), FlagDescTestClass.doubleFlag.get());
-    assertEquals(leftOverArgs.size(), 0);
+    assertEquals(0, leftOverArgs.size());
```
oferb: Why did you switch the order? The expected value is supposed to be first.

common/flags/tests/FlagsTest.java
```diff
@@ -133,14 +134,16 @@ public void firstArgsThenFlagsTest() {
   }
 
   @Test
-  public void printUsageTest() {
+  public void printUsageTest() throws IOException {
```
oferb: How come this compiled before you added `throws IOException`?
I thought if code can throw an exception, it won't compile unless you declare it, as you now did.

common/TextDifferencer.java
val-fed: I meant the real parameters of the method and their description in Javadoc are different.

common/flags/tests/FlagsTest.java
val-fed: It was compiling successfully until I added `try-with-resources` construction for `ByteArrayOutputStream`. After that, I needed to add an exception to the method signature or catch block. Is `try-with-resources` construction redundant for a unit test?

jbduncan: @val-fed 99% of the time, `try-with-resources` is _very_ good to use regardless of whether the code is test code or production code. :)

In this particular case, `try-with-resources` is not strictly necessary, because IIRC, `ByteArrayOutputStream.close()` (which is what gets called by the `try-with-resources` block when the block is exited) doesn't actually do anything, since `ByteArrayOutputStream` doesn't manage any external I/O resources that need to be explicitly closed.

So long story short, in this case, `try-with-resources` is a bit redundant. :)

common/flags/tests/FlagsTest.java
jbduncan: I can't speak for @val-fed, but I believe having the constant or _expected value_ first is the idiomatic way to write JUnit 4 `Assert`-based assertions. The JUnit 4 javadocs seem to confirm this: https://junit.org/junit4/javadoc/4.8/org/junit/Assert.html#assertEquals(java.lang.Object,%20java.lang.Object)

oferb: You're right. I just got confused. @val-fed you can ignore this comment :)
Thanks @jbduncan !

common/flags/tests/FlagsTest.java
oferb: @jbduncan but then any dev (also those just reading the code) should know that about `ByteArrayOutputStream`, otherwise it looks like a mistake.
This was done based on a linter, right @val-fed ?
If so, then perhaps the linter could somehow be smarter about that.

common/TextDifferencer.java
oferb: So please add that to the TODO:
`// TODO Check params description in Javadoc matches actual method params`

PR #138, opened 2018-06-23
CI additional formatting checks

vmax:


jbduncan:
Hmm, I fail to understand why a deployed, binary version of SimpleFormatterTool was introduced (as of commit https://github.com/google/startup-os/pull/138/commits/a0ff865496c61b629bd7a79ba3afbe839e610986) and then the source code for SimpleFormatterTool was in turn apparently removed.

Does this not make SimpleFormatterTool very difficult to build again and audit if/when needed? :thinking:

jbduncan:
Oops, sorry, `SimpleFormatterTool` actually still exists as of commit https://github.com/google/startup-os/pull/138/commits/6f1aecf43ecfe4fc865682975cbd0ddbbc013683. Apologies for the noise! :bowing_man:

check-formatting.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+# Pre-commit hook for StartupOS
+# Run it before committing to verify whether all files 
+# are properly formatted so you won't fail early on review
+# Either do it manually or by creating a symlink
+# To do this, execute, from repo root
+# ln -s $(pwd)/pre-commit.sh $(pwd)/.git/hooks/pre-commit
```
oferb: How about instead of precommit, we add it to `aa fix`?

check-formatting.sh
```diff
@@ -0,0 +1,26 @@
...
+
+
+npm install &>/dev/null
+
+BUILD_FILES=$(find `pwd` -type f \( -name BUILD.bazel -or -name BUILD \) | grep -v node_modules)
+
+bazel run //tools:buildifier -- -mode=check $BUILD_FILES &>/dev/null || (
```
oferb: Can you document what `$(tput setaf 1)[!]` and `$(tput sgr0)` do?

oferb: Can you document what `-mode=check` does?
Also, the `echo` part happens only if `buildifier` fails, right? How does that work?

check-formatting.sh
```diff
@@ -0,0 +1,26 @@
...
+	echo "$(tput setaf 1)[!] BUILD files are not formatted$(tput sgr0)";
+	echo "Please run ''yarn buildifier'' to fix it" 
+	exit 1
+)
+
+bazel run @simple_formatter//:simple_formatter_tool -- --path $(pwd) --java --python --proto --cpp --ignore_directories $(pwd)/node_modules/,$(pwd)/tools/local_server/web_login/node_modules/ &>/dev/null
```
oferb: Could we insert this text into `simple_formatter_tool`, so here we just run it?

tools/aa/commands/InitCommand.java
```diff
@@ -25,13 +25,12 @@
...
-/** A command to init a base folder.
+/**
+ * A command to init a base folder.
  *
- * Usage:
- * bazel run //tools/aa:aa_tool -- init --base_path </path/to/base/folder>
- * or, if aa is already set up:
- * aa init --base_path </path/to/base/folder>
-*/
+ * <p>Usage: bazel run //tools/aa:aa_tool -- init --base_path </path/to/base/folder> or, if aa is
```
oferb: Please make this a block comment and not Javadoc, so we can conserve the formatting:
```
/**
 * Javadoc
 */
```
```
/*
 * Block comment
 */
```

tools/simple_formatter/SimpleFormatterTool.java
```diff
@@ -0,0 +1,199 @@
...
+import java.nio.file.StandardCopyOption;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+class SimpleFormatterTool {
+
+  static String readAllFile(Path path) throws IOException {
```
oferb: How about 
s/readAllFile/readFile

tools/simple_formatter/SimpleFormatterTool.java
```diff
@@ -0,0 +1,199 @@
...
+    try (Stream<Path> stream = Files.walk(Paths.get(path.get()).toAbsolutePath())) {
+      Iterable<Path> paths = () -> stream.filter(Files::isRegularFile).iterator();
+
+      for (Path path : paths) {
+        try {
+          if (shouldFormatFile(path, ignoredDirectories)) {
+            if (dryRun.get()) {
+              System.out.println(
+                  String.format("Intending to format: %s", path.toAbsolutePath().toString()));
+              exit_code = 1;
```
oferb: Why do we return `exit_code = 1`? and shouldn't we then also return 1 when formatting changed the file?

WORKSPACE
```diff
@@ -10,6 +10,13 @@ git_repository(
     remote = "https://github.com/vmax/bazel-deps-deploy",
 )
 
+# deployed version of //tools:simple_formatter
```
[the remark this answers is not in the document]
vmax: @oferb 
>Are we using this?

Yes, [here](https://github.com/google/startup-os/blob/6f1aecf43ecfe4fc865682975cbd0ddbbc013683/check-formatting.sh#L21)

>If so, why not use only //tools:simple_formatter?

Then it would take time to build on `lint` stage, but it's way too much since it depends on `protobuf`

check-formatting.sh
vmax: @oferb yeah, let's remove precommit mentions from both files (`check-formatting.sh` and `pre-commit.sh`)

check-formatting.sh
vmax: @oferb yes. I'll restructure it a bit to be self-documenting.

vmax: @oferb 

>Also, the echo part happens only if buildifier fails, right? How does that work?

Yes, only if `buildifier`'s exit code (`$?`) was non-zero. And `buildifier -mode=check` returns `0` only if all files are properly formatted.

tools/aa/commands/InitCommand.java
vmax: @oferb ok. do you think we can automate this check somehow? we still *do* need `Javadoc` sometimes, right?

tools/simple_formatter/SimpleFormatterTool.java
vmax: @oferb actually, this is a leftover. I first thought that `--dry-run` in my implementation would print out files that would be **changed** by formatting, but in fact it prints out what files it _intends_ to format (without considering it _needs_ formatting). Since then, I've taken a different  approach: to check, we **do** perform the formatting and then check if `git` reports that something changed.

check-formatting.sh
vmax: @oferb we still have to pass `$(pwd)`so let's _not_ do it 🙂
this way we can also migrate same version to `devenv`

check-formatting.sh
```diff
@@ -18,7 +18,7 @@ bazel run //tools:buildifier -- -mode=check $BUILD_FILES &>/dev/null || (
 	exit 1
 )
 
-bazel run @simple_formatter//:simple_formatter_tool -- --path $(pwd) --java --python --proto --cpp --ignore_directories $(pwd)/node_modules/,$(pwd)/tools/local_server/web_login/node_modules/ &>/dev/null
+bazel run //tools/simple_formatter:simple_formatter -- --path $(pwd) --java --python --proto --cpp --ignore_directories $(pwd)/node_modules/,$(pwd)/tools/local_server/web_login/node_modules/ &>/dev/null
```
oferb: You can use this instead:
`//tools/simple_formatter`

PR #140, opened 2018-06-26
Reviewer web app - localserver integration

webdevelopland:
What's new:
- New proto. A lot of code was changed to support it.
- Files are loaded from localserver, instead of using mock-files.
- Firestore (instead of Firebase Realtime database)
- Working with binary (instead of json)
- Working with Diff (proto data) instead of Diff.AsObject.


tools/reviewer/web/src/styles/styles.scss
```diff
@@ -5,9 +5,6 @@
 @import 'global';
 @import 'reviewsPanel';
 @import 'login';
-
-@import 'review/singleReview';
-@import 'review/titlebar';
-
+@import 'review.component.scss';
```
MateMahinKhan: so we have moved scss to files within the component folders?

tools/reviewer/web/src/app/dashboard/reviews-panel/review/person-list/person-list.component.ts
```diff
@@ -1,97 +1,86 @@
...
-      this.onUpdateDiff.emit(this.diff);
-    }
-  }
+    reviewers = reviewers.concat(otherReviewers);
+    this.diff.setReviewerList(reviewers);
 
-  // Save needAttentionOf list
-  saveAttentionList(name: string): void {
-    this.onAddToAttentionList.emit(name);
+    this.firebaseService.updateDiff(this.diff).subscribe(() => {
```
MateMahinKhan: We have the same code at other place as well.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-files/review-files.component.ts
```diff
@@ -1,25 +1,27 @@
...
   // Upon click on a file open a single file review page showing
   // code difference and comments
-  openFile(filePosition): void {
-    // Build a route path on the following format /diff/<diff number>/<path>?
-    // ls=<left snapshot number>&rs=<right snapshot number>
-    this.router.navigate(['diff/' + this.diffId + '/' + filePosition], {
-      queryParams: { ls: '1', rs: '3' }
-    });
+  openFile(filename: string): void {
+    this.router.navigate(['diff/' + this.diffId + '/' + filename]);
```
MateMahinKhan: let's keep the comments for URL building unless Ofer has advised otherwise that I'm not aware of.

MateMahinKhan:
Working with Diff (proto data) instead of Diff.AsObject.
Why did we start using `AsObject` before?

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/code-block/code-block.component.ts
```diff
@@ -1,3 +1,5 @@
+// There's a bug. Lines with more than 100 chars cross borders of code block.
+// TODO: fix the bug
```
MateMahinKhan: You can easily put a scroll there like it was before.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/person-list/person-list.component.ts
```diff
@@ -1,97 +1,86 @@
...
-        break;
-      case 'cc':
-        this.persons = this.diff.getCcList();
-        break;
-      default:
-        throw new Error('Unsupported property');
-    }
+    this.reviewers = this.diff.getReviewerList()
+      .filter(reviewer => {
+        return reviewer.getNeedsattention() === this.isNeedAttention;
```
MateMahinKhan: `reviewer.getNeedsAttention`

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-titlebar/review-titlebar.component.html
```diff
@@ -1,91 +1,37 @@
...
-        </span>
-      </ng-container>
-      <ng-container *ngSwitchCase="'REVERTED'">
-        <span>
-          <mat-icon class="reverted">cancel</mat-icon>
-          Reverted
-        </span>
-      </ng-container>
-    </ng-container>
+    <span>{{ statusList[diff.getStatus()] }}</span>
```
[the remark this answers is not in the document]
webdevelopland: No, it's ok. `statusList` it array, not object. So you can call it by index. Actually, there are no other ways to call it.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-titlebar/review-titlebar.component.ts
```diff
@@ -1,18 +1,53 @@
...
+  }
+
+  updateAuthor(): void {
+    // Update need attention of the author in the diff
+    const author = this.diff.getAuthor();
+    author.setNeedsattention(!author.getNeedsattention());
+    this.diff.setAuthor(author);
+
+    // Send updated diff to firebase
+    this.firebaseService.updateDiff(this.diff).subscribe(() => {
```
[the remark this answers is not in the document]
webdevelopland: updateDiff is already present in firebaseService. All children can use it instead.
Gold rule: If you can put a reusable code to service, then you definitely should do it.
Use component binding for communication between parent and child only.

tools/reviewer/web/src/styles/styles.scss
webdevelopland: Before that we had 3 files:
`reviewsPanel.scss`
`codeReview.scss`
`singleReview.scss`

How should developers know, which file describe review.component?

I renamed the `singleReview` to `review.component.scss`, to understand which scss file describes review.component and its children.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/person-list/person-list.component.ts
webdevelopland: It's fine.
We have freedom to set success and error message in the future or add some pipe directly.
This is much more intuitive, then create additional method.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-files/review-files.component.ts
webdevelopland: There's already comment about it:
https://github.com/google/startup-os/blob/WD1/tools/reviewer/web/src/app/dashboard/reviews-panel/diff/diff.component.ts#L59

tools/reviewer/web/src/app/dashboard/reviews-panel/review/person-list/person-list.component.ts
webdevelopland: No, `getNeedsattention()` it correct

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-files/review-files.component.ts
MateMahinKhan: Sorry, I don't see a comment that says the URL should be of a specific format but maybe it's a discussion between you and Ofer. Anyway, it's of little importance.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/person-list/person-list.component.ts
MateMahinKhan: Sorry, I mean that person-list shouldn't update the diff. Only review should update it and we can avoid calling this service method in a number of components.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-titlebar/review-titlebar.component.ts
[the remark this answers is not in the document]
MateMahinKhan: One of the basic concepts off OOP is separation of concern. review-titlebar.component shouldn't be updating the diff even if it's calling a service method. It should only do stuff a titlebar is supposed to do, that is telling it's parent about a change. It's parent should be updating the diff. That's what I'm asking.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-titlebar/review-titlebar.component.html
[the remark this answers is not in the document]
MateMahinKhan: Alright! I see. What about icons? Can you please restore them?

tools/reviewer/web/src/styles/styles.scss
MateMahinKhan: I didn't ask that but thank you for bringing it up. I agree the names were not proper before but review.component.scss should be review.scss. There is no need for putting component with it. Also you need to specify extension while importing as visible in other imports.

I believe it was a folder of review and two files under the folder. I thought we had decided something else in the chat a couple days ago.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/person-list/person-list.component.ts
MateMahinKhan: It would be better if you could elaborate your answer with some supporting arguments rather than just declaring something. Maybe then I would be better able to see your point.

tools/reviewer/web/src/app/dashboard/index.ts
```diff
@@ -14,7 +14,7 @@ import {
   ReviewDiscussionComponent,
   ReviewFilesComponent,
   ReviewsPanelComponent,
-  ReviewTitlebarComponent
+  ReviewTitlebarComponent,
```
MateMahinKhan: Can we fix the trailing commas so that it's consistent in the whole app. I don't care if are enforcing trailing comma or enforcing no-trailing comma. But it has to be consistent in the whole app at any given point in time.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/person-list/person-list.component.ts
webdevelopland: `getNeedsattention()` is an method from proto function.
It's autogenerated. I don't choose a case of the names.
Original code:
```typescript
export class Reviewer extends jspb.Message {
  getEmail(): string;
  setEmail(value: string): void;

  getNeedsattention(): boolean;
  setNeedsattention(value: boolean): void;
```
You can see it by yourself by run 'npm run protoc` and open
`src/app/shared/shell/proto`

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-files/review-files.component.ts
webdevelopland: I don't see any necessity to add something else to url or maintain some specific url format.
Maybe we'll need to add commit data to it in the future. Exactly the question is raised in the link, which I send above.

MateMahinKhan: We'll definitely need to add commit data to it in the future and in the format described in the comment. That is exactly why we need to maintain a specific url format.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/person-list/person-list.component.ts
MateMahinKhan: bro, I didn't say that there's an error. Yeah, I said to change it as it is the simplest thing to say as it looked like the casing was incorrect. As I said It would have been better if you gave this explanation in your first response rather than declaring that the casing is correct and saved both of us time. It is not possible for a reviewer to know everything about code.

tools/reviewer/web/src/app/dashboard/index.ts
MateMahinKhan: Yeah, let's do that.

oferb: It's great that you agree.
Do you know of a tool that can automatically check this?

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/diff.component.ts
```diff
@@ -1,14 +1,26 @@
+// NOTICE: it's actually not a diff, it's a two files changes
```
oferb: > it's actually not a diff, it's a two files changes
What do you mean? Isn't a diff the changes between 2 files?

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/diff.component.ts
```diff
@@ -1,14 +1,26 @@
...
+// TODO: rename the component and linked files
```
oferb: Rename to what?

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/diff.component.ts
```diff
@@ -63,24 +102,51 @@ export class DiffComponent implements OnInit, OnDestroy {
...
+    this.firebaseService.updateDiff(this.diff).subscribe();
+  }
+
+  createNewThread(lineNumber: number, comments: Comment[]): Thread {
+    const newThread = new Thread();
+    newThread.setLineNumber(lineNumber);
+    newThread.setIsDone(false);
+    newThread.setCommentList(comments);
+
+    // Temperoally hardcoded data
```
oferb: s/Temperoally/Temporarily

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/diff.component.ts
```diff
@@ -17,42 +29,69 @@ import { files } from './mock-files';
...
-      newThread.setIsDone(false);
-      newThread.setCommentsList(comments);
-      this.diff.addThreads(newThread);
-    }
+  getFileChanges(): void {
+    const textDiffRequest = new TextDiffRequest();
 
-    this.firebaseService.updateDiff(this.diff).subscribe();
+    // Temperoally hardcoded data
+    // TODO: take the data from somewhere else (url?)
```
oferb: Is this about how to get the data from a different part of the app?
Maybe @MateMahinKhan has some ideas.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/person-list/person-list.component.ts
```diff
@@ -1,97 +1,101 @@
...
+    reviewers = reviewers.concat(otherReviewers);
+    this.diff.setReviewerList(reviewers);
+
+    this.reviewService.saveLocalDiff(this.diff);
   }
 
-  // Save needAttentionOf list
-  saveAttentionList(name: string): void {
-    this.onAddToAttentionList.emit(name);
+  // Add to CC, if reviewer is inculed in Reviewers.
```
oferb: s/inculed/included

tools/reviewer/web/src/app/dashboard/reviews-panel/review/person-list/person-list.component.ts
```diff
@@ -1,97 +1,101 @@
...
+  // Add to Reviewers, if reviewer is inculed in CC.
+  changeAttentionOfReviewer(reviewer: Reviewer): void {
```
oferb: I think this needs to be changed - let me explain the concepts better.


CC is never in the "Attention set".
CC is a passive observer.

Only Author and Reviewers can be in the Attention Set.

Attention Set is pretty much - who's turn is it to do something with this Diff. Sometimes it's the Author, sometimes it's the Reviewer. It can also be both. It can also be only some of the reviewers but not all.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-files/review-files.component.ts
oferb: Sorry, I don't follow.. 
I'm fine with the code as it is. Vadim removed a comment that explained the `this.router.navigate()` call. I think that's fine.
There wasn't any discussion around this AFAIR, and the format can change. It just needs to be reasonable and have all that's needed to be able to send the url to someone and he should open the same page.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-titlebar/review-titlebar.component.html
```diff
@@ -1,93 +1,42 @@
...
-        (click)="saveAttentionList(diff.author);
-        $event.stopPropagation()"
-        >
+        (click)="updateAuthor()"
+      >
         {{ 
-          diff.needAttentionOfList.indexOf(diff.author) > -1
-          ? 'Remove from Needs Attention'
-          : 'Add to Needs Attention'
+          diff.getAuthor().getNeedsattention()
```
oferb: s/getNeedsattention/getNeedsAttention

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-titlebar/review-titlebar.component.ts
```diff
@@ -1,18 +1,42 @@
...
+  statusList = statusList;
+  isLoading: boolean = true;
+  @Input() diff: Diff;
   @Input() editable;
-  @Output() onAddToAttentionList = new EventEmitter<string>();
 
-  // Save needAttentionOf list
-  saveAttentionList(name: string): void {
-    this.onAddToAttentionList.emit(name);
+  // Ask review.component to send diff to firebase
```
oferb: What's review.component?
I see that the diff is saved in `this.reviewService.saveLocalDiff()`

tools/reviewer/web/src/app/dashboard/reviews-panel/reviews-panel.component.ts
```diff
@@ -1,70 +1,67 @@
...
-            this.diffs[Lists.CcedOn].push(diff);
-          } else if (diff.author === this.username && diff.status === 0) {
-            // Draft Review
-            this.diffs[Lists.DraftReviews].push(diff);
-          } else if (diff.author === this.username && diff.status === 4) {
-            // Submitted Review
-            this.diffs[Lists.SubmittedReviews].push(diff);
+        for (const diff of diffs) {
+          for (const reviewer of diff.getReviewerList()) {
+            if (reviewer.getEmail() === this.authService.userEmail) {
```
oferb: Looks like you have the same condition twice:
`reviewer.getEmail() === this.authService.userEmail`

tools/reviewer/web/src/app/shared/services/auth.guard.ts
```diff
@@ -10,19 +10,23 @@ import { Observable } from 'rxjs';
...
+  constructor(private router: Router, private authService: AuthService) { }
 
   canActivate(
     route: ActivatedRouteSnapshot,
     state: RouterStateSnapshot
   ): Observable<boolean> {
+    // There is a bug.
+    // If you open the app without internet connection,
+    // you don't see any loadings or error message.
+    // You see nothing, just white screen forever.
```
oferb: LOL :-)

tools/reviewer/web/src/assets/code-review.proto
```diff
@@ -1,60 +1,168 @@
 syntax = "proto3";
-package messages;
 
+message File {
```
oferb: Please add the comment from Slack about this file, and a TODO to fix it.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-titlebar/review-titlebar.component.ts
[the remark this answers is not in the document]
oferb: @MateMahinKhan where do you see identical code at many places?
Also, my personal opinion about OOP concepts - they sometimes lead to more complicated code. There's some important properties of code (correctness, performance, simplicity, consistent style, maintainability, testability etc). These concepts are important also for non-OOP code, and IMO are more important than adhering to OOP concepts.

In this case, if the review-titlebar.component needs to update the diff, but it shouldn't (to adhere to a basic OOP concept), and instead it should call the parent, which will have the method that does this, then - 
* If we just call the method `updateDiff()` we aren't really doing much except some method indirection. It's pretty clear to anyone reading the code that Titlebar is updating the diff.
* If we don't call it `updateDiff()` (or similar), then it needs to be some abstraction of the action taking place. For example - `handleAuthorUpdate()` (but then you might need also `handleReviewerUpdate()`). So maybe more abstract - `handleTitleBarAction()` or something like that, so the Titlebar doesn't know this, but the diff is actually being saved. To me, although this is correct in one sense, Titlebar doesn't deal with saving Diff directly, it makes the code less simple, while not providing anything very useful for the other properties (correctness, performance etc).

So, in summary, IMO concepts like separation of concern, DRY and others (e.g http://java-design-patterns.com/principles), are all kind of ok, but can sometimes lead you to code that's actually worse in some aspects, and not better (or marginally better) in others.

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/diff.component.ts
webdevelopland: We have `Diff` in out proto. It includes all files, changes, comment etc. But the component doesn't describe whole `Diff`, it describes one file only (and its changes)

webdevelopland: We have review.component, which actually describe `Diff`

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/diff.component.ts
webdevelopland: e.g. `FileChangeComponent`

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/diff.component.ts
webdevelopland: No, it's not about it.
It's part of communication with local server, which I don't understand.
In this part of the app we have
1) Diff from firebase
2) filename
How we're going to know, which `commit_id` should be used?

webdevelopland: Maybe we should put the commit ids to url?

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-titlebar/review-titlebar.component.html
webdevelopland: :)

tools/reviewer/web/src/app/dashboard/index.ts
MateMahinKhan: Yeah, we are using linter already.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-files/review-files.component.ts
MateMahinKhan: @oferb we talked about the URL will have a format including leftsnapshot and rightsnapshot number. We're discussing that comment. It is relevant to this discussion https://github.com/google/startup-os/pull/140/files#r198593942

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-titlebar/review-titlebar.component.html
MateMahinKhan: 😆  @oferb join the club!

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-files/review-files.component.ts
webdevelopland: We don't use snapshots anymore.

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-titlebar/review-titlebar.component.ts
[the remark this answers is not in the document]
MateMahinKhan: @oferb It was a method of updating the diff. 
Yeah, I'm of the strong opinion that titlebar shouldn't know about saving the diff. Angular 2 has made it really easy to make components and separate code into concerns and follow  I'm not sure but I guess it will result in better testable code as well. I see your point that it makes code complicated. 

Vadim, recently, started using review service that does the job as well of only one point of updating the diff from any child component.

tools/reviewer/web/src/app/shared/services/auth.guard.ts
webdevelopland: :-)

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-titlebar/review-titlebar.component.ts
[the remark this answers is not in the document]
oferb: @MateMahinKhan I agree that Angular has some pretty good architecture guidelines for separating things correctly. All I'm saying is that we can choose whether to do that, and in this case, I don't think it will bring much value, since we aren't really testing the code.

As for tests - refactoring his code later to call the parent component would be fairly simple, so we can do that when we need to, for tests. Until then - we get simpler code.

Because we can refactor later, I view this as a reversible decision (AKA type 2 decision http://www.businessinsider.com/jeff-bezos-on-type-1-and-type-2-decisions-2016-4). There are decisions on testability that are type 1 decisions, like choosing a framework that is more testable (AFAIU Angular is more testable than React).

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review.component.html
```diff
@@ -18,34 +16,28 @@
...
+          ></app-person-list>
         </div>
   
         <!-- Person List based on CC in the Diff -->
         <div>
           <app-person-list 
             [diff]="diff"
-            [property]="'cc'"
-            [editable]="editable"
+            [isNeedAttention]="false"
```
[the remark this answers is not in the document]
oferb: I don't follow. I know in the previous code it was the same code if an if statement inside for `cc` or `reviewer`. How does it work now without `property`?

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-files/review-files.component.ts
oferb: I forgot about that! So yes, we should have those parameters in the url, but they should point to commits, not snapshots. They could still be a number (doesn't have to be the commit sha).

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/diff.component.ts
oferb: So maybe rename to `DiffFile`?

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/diff.component.ts
oferb: Good. Can you add that as a suggestion to the TODO?

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-titlebar/review-titlebar.component.html
oferb: :-)
LOL
GMTA

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/diff.component.ts
oferb: So the commit_id should be the last one on the right.
On the left, let's put the first commit for now. Later we will put the last-reviewed one, but we need to mark it as such, which we don't currently.

tools/reviewer/web/src/app/dashboard/reviews-panel/reviews-panel.component.ts
```diff
@@ -39,23 +35,34 @@ export class ReviewsPanelComponent implements OnInit {
...
-                this.diffGroups[Lists.NeedAttention].push(diff.toObject());
-              } else {
-                // User is cc'ed on this
-                this.diffGroups[Lists.CcedOn].push(diff.toObject());
-              }
-            } else if (reviewer.getEmail() === this.authService.userEmail) {
-              if (diff.getStatus() === Diff.Status.REVIEW_NOT_STARTED) {
+          diff.getDescription();
+          ///
+          const you: string = this.authService.userEmail;
```
oferb: you is you or me? or me is you?
How about we call it `user` or `currentUser`?
I have a friend who wrote a server that printed "My bad" and "Your bad" on errors when it was the server's fault or the caller's fault. It was funny but kinda confusing.

tools/reviewer/web/src/assets/code-review.proto
```diff
@@ -135,14 +139,15 @@ message Diff {
...
   repeated Reviewer reviewer = 3;
-  string description = 4;
-  string bug = 5;
-  Status status = 6;
-
-  string workspace = 7;
-  int64 created_timestamp = 8;
-  int64 modified_timestamp = 9;
-  repeated Thread thread = 10;
+  repeated string cc = 4; // email list
```
oferb: You don't have to change the code, but just know that any binary proto that you try to load, assumes that you're loading it with the same IDs with which it was created.
So for example, `string bug = 5;` changed to `string description = 5;` so if you try to load old binary data, field #5 with bug data will be put into the description.
In many cases though, you won't be able to read the data at all, since the types changed - i.e try to load a string into and int.
Right now it's alright since we can delete all data in the db anyway. Once we start to have meaningful data, we can't change it anymore.

tools/reviewer/web/src/app/dashboard/reviews-panel/reviews-panel.component.ts
```diff
@@ -39,23 +35,34 @@ export class ReviewsPanelComponent implements OnInit {
...
+          } else if (diff.getCcList().includes(you)) {
+            // User is cc'ed on this
+            this.diffGroups[Lists.CcedOn].push(diff);
+          } else if (diff.getAuthor().getEmail() === you) {
+            switch (diff.getStatus()) {
+              case Diff.Status.REVIEW_NOT_STARTED:
                 // Draft Review
-                this.diffGroups[Lists.NeedAttention].push(diff.toObject());
-              } else if (diff.getStatus() === Diff.Status.SUBMITTED) {
+                this.diffGroups[Lists.NeedAttention].push(diff);
```
oferb: I know it's not your change, but - 
This one should go into `DraftReviews`

tools/reviewer/web/src/app/dashboard/reviews-panel/reviews-panel.component.ts
```diff
@@ -39,23 +35,34 @@ export class ReviewsPanelComponent implements OnInit {
...
+                break;
+              case Diff.Status.SUBMITTED:
                 // Submitted Review
-                this.diffGroups[Lists.CcedOn].push(diff.toObject());
-              }
+                this.diffGroups[Lists.CcedOn].push(diff);
```
oferb: Also, not your change, but:
* This one should go to `SubmittedReviews`
* `SubmittedReviews` should be first, so if it's already submitted, it should not go into the other lists.

tools/reviewer/web/src/app/dashboard/reviews-panel/reviews-panel.component.ts
webdevelopland: xD

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review.component.html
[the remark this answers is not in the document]
webdevelopland: @oferb it has `property` now.

tools/reviewer/web/src/app/dashboard/reviews-panel/diff/diff.component.ts
webdevelopland: @oferb already did

tools/reviewer/web/src/app/dashboard/reviews-panel/review/review-titlebar/review-titlebar.component.html
webdevelopland: @oferb oh, wow. It's a mistake in proto. Camel case instead of snake case.
Cool. Yeah, I'll leave a comment

tools/reviewer/web/src/assets/code-review.proto
webdevelopland: Ok, I'll be aware of it.
Actually, it's a relief. It was really annoying to change the indexes all the time.

tools/reviewer/web/src/app/dashboard/reviews-panel/reviews-panel.component.ts
```diff
@@ -1,88 +1,95 @@
...
+            this.diffGroups[Lists.CcedOn].push(diff);
+          } else if (diff.getAuthor().getEmail() === userEmail) {
+            switch (diff.getStatus()) {
+              case Diff.Status.REVIEW_NOT_STARTED:
+                // Draft Review
+                this.diffGroups[Lists.DraftReviews].push(diff);
+                break;
+              case Diff.Status.SUBMITTED:
+                // Submitted Review
+                this.diffGroups[Lists.SubmittedReviews].push(diff);
```
oferb: `SubmittedReviews` should be first, so if it's already submitted, it should not go into the other lists.

PR #143, opened 2018-07-04
Fixup aa tool

vmax:


tools/local_server/run.sh
```diff
@@ -15,9 +16,12 @@ fi
 # Kill previous:
 if [ "$(uname)" = "Darwin" ] || [ "$(uname)" = "Linux" ]; then
   # Mac & Linux
-  pkill -f 'tools/local_server/local_http_gateway'
-  pkill -f 'tools/local_server/local_server'
-  kill $(lsof -t -i:8000) # For Angular
+  # kill gateway (port 7000), gRPC server (port 8001) and Angular server (8000)
```
oferb: s/gateway/HTTP gateway

tools/local_server/run.sh
```diff
@@ -15,9 +16,12 @@ fi
...
+  # -t makes `lsof` output only PIDs so output can be piped to `kill`
+  # -n and -P prevent `lsof` from resolving addresses and ports, therefore
+  # making execution faster
```
oferb: Please mention that we filter by TCP:LISTEN to avoid killing Chrome, which is also on 8000.

PR #144, opened 2018-07-05
Reviewer - proto import fix

webdevelopland:
What's new:
- Web app uses original proto, instead of a copy
- original proto is changed. id, needs_attention, email, etc...
- gulp script, which creates proto function is updated
- Few changed in the code, to support new proto


tools/reviewer/web/gulpfile.js
```diff
@@ -1,69 +1,96 @@
...
+const clean = require('gulp-clean');
+const mv = require('mv');
 const { exec } = require('child_process');
 
-const pathToProto = './src/assets';
+const tempPath = './temp';
 const shellPath = './src/app/shared/shell/proto';
-const protoName = 'code-review';
-const protoPath = path.join(pathToProto, protoName + '.proto');
+const codeReviewProtoPathRelative = 'tools/reviewer/service/code_review.proto';
```
oferb: s/PathRelative/RelativePath

tools/reviewer/web/gulpfile.js
```diff
@@ -1,69 +1,96 @@
...
+// from 'startup-os/tools/reviewer/web' to 'startup-os'
+const startuposPath = path.resolve('../../../');
 
 gulp.task('default', ['protoc']);
 
 gulp.task('protoc', () => {
```
oferb: Can you give a general overview of the `protoc` gulp task - why we're doing it like we're doing it?

tools/reviewer/web/gulpfile.js
```diff
@@ -1,69 +1,96 @@
...
+    codeReviewProtoPathRelative
+  );
+  var protoImportList = [];
+  const codeReviewFilename = path.parse(codeReviewProtoPathRelative).base;
+  protoImportList.push(codeReviewFilename);
+  const newCodeReviewPath = path.join(tempPath, codeReviewFilename);
+  fs.copySync(codeReviewProtoPath, newCodeReviewPath);
+  const codeReviewContent = fs.readFileSync(newCodeReviewPath, 'utf8');
+  const codeReviewLines = codeReviewContent.split('\n');
+  for(let line of codeReviewLines) {
```
oferb: Nit: Add a space (`for (`)

tools/reviewer/web/gulpfile.js
webdevelopland: 1) protoc saves path of input in out output..
For example.
If your input: `./aaa/bbb/data.proto`
and you set output as: `./zzz`
protoc create a function here: `./zzz/aaa/bbb/function.js`
That is why we create temp folder in a root of web reviewer.

2) We can't use a parent path with protoc. Such as '../../../'
So if we want to get access to proto file, which is located not inside `startup-os/tools/reviewer/web` then we have to run protoc from root `./startup-os`.
I think that's not what we need. I tried to keep all manipulation with web app inside web app project.

3) protoc don't create 1 proto function from 3 proto files. It creates also 3 functions. (Actually even 6, 1 js and 1 ts on each 1 proto file)

4) proto files have exact paths.
e.g. `common/repo/repo.proto`
If we don't want to edit proto files, then we have to keep that structure. That's why I didn't put all proto function in one folder.

oferb: Great :)
So can you just add these explanations to the code?

PR #145, opened 2018-07-06
Updated CI and fix formatting

vmax:


WORKSPACE
```diff
@@ -10,6 +10,12 @@ git_repository(
     remote = "https://github.com/oferb/startupos-binaries",
 )
 
+git_repository(
+    name = "startupos_binaries",
+    commit = "e4ada206e86f94ef78176fa6e78e0d25919ecf7a",
+    remote = "https://github.com/vmax/startupos-binaries",
```
oferb: Please use my repo instead.

check-formatting.sh
```diff
@@ -9,16 +9,17 @@ RESET=$(tput sgr0)
 
 npm install &>/dev/null
 
+PROTO_OPTIONS="--proto_compiler //external:proto_compiler --proto_toolchain_for_java //external:proto_java_toolchain"
```
oferb: Do we still need these if we add them to `bazelrc`?

(throughout this PR)

compile.sh
```diff
@@ -0,0 +1,21 @@
...
+# Run with `bazel` command as param (build|test)
+if [[ $1 != "build" && $1 != "test" ]]; then
+  echo "$RED""Run script with 'build' or 'test' as param$RESET"
+  exit 1
+fi
+
+if [[ ! -z "$CIRCLECI" ]]; then
+  DELETED_PACKAGES="";
+else
+  DELETED_PACKAGES="--deleted_packages $(cat .circleci/deleted_bazel_packages.txt)";
```
oferb: Can you document what `--deleted_packages` are and why they're needed?

WORKSPACE
vmax: @oferb I'll be able to if you downgrade the grpc plugin version to `v0.12.0` -- otherwise build will fail

check-formatting.sh
vmax: @oferb I think we don't

compile.sh
vmax: @oferb done

compile.sh
```diff
@@ -0,0 +1,23 @@
...
+# Run with `bazel` command as param (build|test)
+if [[ $1 != "build" && $1 != "test" ]]; then
+  echo "$RED""Run script with 'build' or 'test' as param$RESET"
+  exit 1
+fi
+
+# On CircleCI we ignore packages that depend on
+# @com_google_protobuf//:protobuf (cpp library),
+# which leads to inability to use prebuilt binaries
+# and significantly increases build time
```
oferb: Can you please add a TODO to be able to compile C++ using prebuilt binaries?

third_party/java_grpc_library.bzl
```diff
@@ -0,0 +1,134 @@
+# SRC: https://github.com/grpc/grpc-java/blob/master/java_grpc_library.bzl
```
oferb: s/SRC/Source

WORKSPACE
```diff
@@ -233,3 +239,33 @@ http_jar(
     sha256 = "c2a453921632c0c3559f9df92e1699b69c784181f36a316f9927b70f52e5a7d5",
     url = "https://github.com/grpc-ecosystem/polyglot/releases/download/v1.6.0/polyglot.jar"
 )
+
+http_file(
+    name = "protoc_bin",
+    executable = True,
+    sha256 = "84e29b25de6896c6c4b22067fb79472dac13cf54240a7a210ef1cac623f5231d",
```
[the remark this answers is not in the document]
vmax: @oferb is there another way? I mean, there are no SHA256s [on download page](https://github.com/google/protobuf/releases)

.circleci/deleted_bazel_packages.txt
```diff
@@ -0,0 +1 @@
+tools/clang_json
```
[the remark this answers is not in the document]
vmax: @oferb it does not help much because `--proto_compiler` and `--proto_toolchain_for_java` options are not available in `query` [is this an issue to create in `google/protobuf`?] and when we don't specify them, `bazel` will implicitly make a dependency i.e. between each `proto_library` rule and `@com_google_protobuf//:protoc`

compile.sh
vmax: @oferb will *we* prebuild those deps and include in `startupos_binaries`?

oferb: Yes

WORKSPACE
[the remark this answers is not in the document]
oferb: I think yours is the only way, just checking :)

PR #154, opened 2018-07-10
Ignore symbolic links during directory copying

vmax:


common/FileUtils.java
```diff
@@ -257,6 +257,9 @@ public FileVisitResult visitFile(final Path file, final BasicFileAttributes attr
                 return FileVisitResult.CONTINUE;
               }
             }
+            if (Files.isSymbolicLink(file)) {
```
oferb: Can you check that it indeed creates the target directory, and then accordingly, add to the method:
`/** Copies a directory to another directory. Creates target directory if needed and doesn't copy symlinks. */`

vmax: @oferb 

>it indeed creates the target directory

https://github.com/google/startup-os/blob/00071441a87426d1e24643688e2927800e5c20c7/common/FileUtils.java#L248

PR #155, opened 2018-07-10
Aa: allow compiling from another workspace

vmax:


tools/aa/aa_tool.sh
```diff
@@ -139,16 +139,18 @@ function aa {
   # Uncomment to override StartupOS repo:
   #STARTUP_OS=<repo path>
   # Uncomment to force recompile:
-  #AA_FORCE_COMPILE=1
+  #AA_FORCE_COMPILE_WS=<workspace_name>
 
   AA_BINARY="$STARTUP_OS/bazel-bin/tools/aa/aa_tool"
-  if [ ! -f $AA_BINARY ] || [ "$AA_FORCE_COMPILE" = "1" ]; then
-    cd $STARTUP_OS
+  if [ ! -f $AA_BINARY ] || [ ! -z "$AA_FORCE_COMPILE_WS" ]; then
```
oferb: This `if` also takes care of the first (normal) compilation of `aa_tool` from `$STARTUP_OS`, but now it always compiles from `$AA_FORCE_COMPILE_WS`.

I think we might need to split this into an (bash-like pseudo-code ahead):
`if [ ! -z "$AA_FORCE_COMPILE_WS" ]; then
...
else
  if [ ! -f $AA_BINARY ]; then
  ...
  fi
fi

PR #170, opened 2018-07-11
Add last master commit to getCommits(), fallback to head if workspace is not found

oferb:


common/repo/GitRepo.java
```diff
@@ -150,15 +150,21 @@ public void tagHead(String name) {
...
+    // We reverse to return by chronological order
+    return splitLines(commandResult.stdout).reverse();
   }
 
   public ImmutableList<Commit> getCommits(String branch) {
     ImmutableList<String> commits = getCommitIds(branch);
     ImmutableList.Builder<Commit> result = ImmutableList.builder();
     for (String commit : commits) {
-      result.add(Commit.newBuilder().setId(commit).addAllFile(getFilesInCommit(commit)).build());
+      if (commit.equals(commits.get(0))) {
```
vmax-reef: @oferb you call `commits.get(0)` on each iteration but the result is constant. What's the performance of `.get()` for `ImmutableList`? Just interested, no need to change

vmax: @oferb you call `commits.get(0)` on each iteration but the result is constant. What's the performance of `.get()` for `ImmutableList`? Just interested, no need to change

oferb: If Java knows this is immutable, it might cache the call.
Otherwise, if `ImmutableList` is backed by an array, it's O(1). If it's backed by a linked-list it's still O(1) because it's the first element (any constant index would be the same). If it's backed by another data structure - it depends.
As for actual runtime performance, I don't know but Java is pretty fast and there typically aren't too many commits in a branch, so it shouldn't be an issue.

Issue #162, opened 2018-07-11
Run `buildifier` as part of multi-language formatter


oferb:
Instead of running buildifier as a separate formatter [here](https://github.com/google/startup-os/blob/master/check-formatting.sh), run it from [SimpleFormatter](https://github.com/google/startup-os/tree/master/tools/simple_formatter).

vmax:
Should have been closed by commit message 🤔

PR #172, opened 2018-07-12
Run buildifier as part of FormatterTool; fix #159, #162

### LINKED ISSUE #159: Rename `SimpleFormatterTool` to `FormatterTool`

vmax:


formatting.sh
```diff
@@ -0,0 +1,21 @@
...
+npm install &>/dev/null
+
+bazel run //tools/formatter -- \
+					--path $(pwd) \
+					--java --python --proto --cpp --build \
+					--ignore_directories $(find $(pwd) -name node_modules -type d | paste -s -d , -) \
+					&>/dev/null
+if [[ ! -z $(git status -s) ]]; then
+	echo "$RED[!] Source files are not formatted$RESET";
+	echo "Please run ''./formatting.sh'' to fix it"
```
oferb: You merged `check` and `fix`?
So if it's called `formatting.sh` now, why would I run it again if I just ran it? (`Please run ''./formatting.sh'' to fix it`)

tools/formatter/FormatterTool.java
```diff
@@ -96,6 +96,17 @@ public void format(Path path) throws IOException {
...
   }
 
+  static class BuildFormatter implements BaseFormatter {
+    static BuildFormatter init() {
+      return new BuildFormatter();
+    }
+
+    @Override
+    public void format(Path path) throws IOException {
+      execute("buildifier -mode=fix " + path.toAbsolutePath().toString());
```
oferb: If we run it as `bazel run //tools:buildifier`, then we don't need to install it, no?

formatting.sh
vmax: @oferb I think it'd be reasonable to either take `check`|`fix` as cmdline argument or to always fix and print out error if we're on CI

tools/formatter/FormatterTool.java
vmax: @oferb we'll have to run it as `bazel run --direct_run`, I think. Let's try

vmax: @oferb actually, if we do it like this, we'll have an implicit dependency on having `buildifier` being available as exact Bazel target `//tools:buildifier` -- but `formatter` is self-contained (except for external `yapf`/`clang-format`/`buildifier` deps). Also, `angular/ngcontainer` *already* has `buildifier` available -- so let's not do it 🙂

oferb: Maybe we can make the dependency explicit using `https://docs.bazel.build/versions/master/be/java.html#java_binary.runtime_deps`?
If we rely on `buildifier` being available, that includes all dev machines, not only `angular/ngcontainer`. Otherwise, devs can't run it the same way.

formatting.sh
oferb: Let's always fix then, but maybe add in a comment how to make it just check.

formatting.sh
```diff
@@ -0,0 +1,21 @@
...
+RESET=$(tput sgr0)
+
+npm install &>/dev/null
+
+bazel run //tools/formatter -- \
+					--path $(pwd) \
+					--java --python --proto --cpp --build \
+					--ignore_directories $(find $(pwd) -name node_modules -type d | paste -s -d , -) \
+					&>/dev/null
+if [[ ! -z "$CIRCLECI" && ! -z $(git status -s) ]]; then
```
oferb: `git status -s` means we're checking modified unstaged files?
What if I'm currently working and I have some modified files?
Also, if I run this twice, will it print the error on the second time?

tools/formatter/FormatterTool.java
```diff
@@ -147,21 +166,25 @@ private static boolean shouldFormatFile(Path file, Set<Path> ignoredDirectories)
...
-          ".proto", ClangFormatter.init(),
-          ".py", PythonFormatter.init(),
-          ".cc", ClangFormatter.init());
+      ImmutableMap.<String, BaseFormatter>builder()
+          .put(".java", JavaFormatter.init())
+          .put(".proto", ClangFormatter.init())
+          .put(".py", PythonFormatter.init())
+          .put(".cc", ClangFormatter.init())
+          .put(".bazel", BuildFormatter.init())
+          .put("", BuildFormatter.init())
```
oferb: No extension it formatted as a BUILD file?

vmax: @oferb 
If file does not have an extension but filename is not equal to `BUILD`, it would *not* be formatted.
It's the limitation of having extensions mapped to formatters (`ImmutableMap<String[extension], BaseFormatter>`). A refactor we could do is change it to `ImmutableMap<Pattern, BaseFormatter`) but then performance drops due to matching filenames via regex.

formatting.sh
vmax: @oferb 
>git status -s means we're checking modified unstaged files?

Right

>What if I'm currently working and I have some modified files?

Then you want fall through `! -z "$CIRCLECI"` condition -- since you're *not* on CircleCI

>Also, if I run this twice, will it print the error on the second time?

It prints out an error only if both conditions are satisfied:
* we are on CircleCI
* working tree contains unstaged changes

When you run it locally it silently fixes everything.

.circleci/config.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       - restore_cache:
           key: "v1-bazel_cache"
       - run: './pre-commit.sh'
-      - run: './check-formatting.sh'
+      - run: './formatting.sh'
```
oferb: How about we call this `fix_formatting.sh`?

tools/formatter/FormatterTool.java
```diff
@@ -96,6 +96,19 @@ public void format(Path path) throws IOException {
     }
   }
 
+  static class BuildFormatter implements BaseFormatter {
+    static BuildFormatter init() {
```
oferb: Why did we do `init()` instead of an empty constructor?

tools/formatter/FormatterTool.java
```diff
@@ -96,6 +96,19 @@ public void format(Path path) throws IOException {
...
+      return new BuildFormatter();
+    }
+
+    @Override
+    public void format(Path path) throws IOException {
+      String command =
+          "/usr/bin/env bash " + "tools/buildtools_wrappers/buildifier.sh -mode=fix " + "%s";
```
oferb: Does this work on both Linux and Mac?

tools/formatter/FormatterTool.java
oferb: How about a refactor (in a separate PR) to:
- Remove `shouldFormatFile()`
- Remove `formatters`
- Instead, have a method `getFormatter()` with the same inputs as `shouldFormatFile`
- Use this input in main loop, instead of `shouldFormatFile`

This would remove the duplication we have now where we have e.g `.java` both in `isJava()` and in `formatters`.

formatting.sh
```diff
@@ -0,0 +1,21 @@
...
+# are properly formatted so you won't fail early on review
+# Execute from repo root or, if using aa from base/head/startup-os
+
+RED=$(tput setaf 1)
+RESET=$(tput sgr0)
+
+npm install &>/dev/null
+
+bazel run //tools/formatter -- \
+					--path $(pwd) \
```
oferb: Looks like there's too much indentation here.

formatting.sh
oferb: Great. Can you add that as a comment just before the `if`?

tools/formatter/FormatterTool.java
vmax: @oferb yes, because script picks the correct binary

tools/formatter/FormatterTool.java
vmax: @oferb Indeed! 🤔

fix_formatting.sh
```diff
@@ -10,10 +10,15 @@ RESET=$(tput sgr0)
...
-					&>/dev/null
+	--path $(pwd) \
+	--java --python --proto --cpp --build \
+	--ignore_directories $(find $(pwd) -name node_modules -type d | paste -s -d , -) \
+	&>/dev/null
+
+# Prints out an error only if both conditions are satisfied:
+# * we are on CircleCI
+# * working tree contains unstaged changes
+# When you run it locally it silently fixes everything.
```
oferb: s/When you/When

tools/formatter/FormatterTool.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.google.startupos.tools.simple_formatter;
+package com.google.startupos.tools.formatter;
```
oferb: Please add a TODO for the comment about the refactor.

PR #174, opened 2018-07-17
Transition from jgit to git

val-fed:


common/repo/repo.proto
```diff
@@ -29,6 +29,7 @@ message File {
     RENAME = 2;
     MODIFY = 3;
     COPY = 4;
+    UNTRACKED = 5;
```
oferb: Let's not add UNTRACKED, and instead use the other ones (e.g ADD).
I understand that ADD can be tracked or untracked, but that distinction is not important at this point.

common/repo/GitRepo.java
```diff
@@ -98,7 +67,26 @@ private CommandResult runCommand(String command) {
     CommandResult result = new CommandResult();
     try {
       List<String> fullCommand = new ArrayList<>(gitCommandBase);
-      fullCommand.addAll(Arrays.asList(command.split(" ")));
+      if (!command.contains("\"")) {
```
oferb: Why did you need all of this new code?

common/repo/GitRepo.java
```diff
@@ -169,42 +156,25 @@ public void tagHead(String name) {
...
-    } catch (GitAPIException e) {
-      throw new RuntimeException(e);
+    CommandResult commandResult = runCommand("status --short");
+    ImmutableList<String> lines = splitLines(commandResult.stdout);
+    for (String line : lines) {
+      String[] parts;
+      if (!line.trim().startsWith("M")
+          || !line.trim().startsWith("RM")
+          || !line.trim().startsWith("??")) {
+        line = line.replaceFirst("\\s{2}", " ");
```
oferb: Can you put `"\\s{2}"` in a variable and name that variable to explain what the regex does?

common/repo/GitRepo.java
```diff
@@ -248,107 +225,73 @@ public void tagHead(String name) {
...
-      // reset to `ref` and all changes introduced after it
-      // would be marked as unstaged but saved in working tree
-      jGit.reset().setMode(ResetType.MIXED).setRef(ref).call();
-    } catch (GitAPIException e) {
-      throw new RuntimeException(e);
-    }
+    // HEAD pointer would be
+    // reset to `ref` and all changes introduced after it
+    // would be marked as unstaged but saved in working tree
+    runCommand("reset " + ref);
```
oferb: We don't need it mixed anymore?
@vmax WDYT?

common/repo/GitRepo.java
```diff
@@ -248,107 +225,73 @@ public void tagHead(String name) {
...
+    CommandResult commandResult = runCommand("merge " + branch + " -s recursive -Xtheirs");
+    String err = commandResult.stderr;
+    return err.length() == 0;
+  }
+
+  public boolean merge(String branch, boolean remote) {
+    switchToMasterBranch();
+    CommandResult commandResult;
+    if (remote) {
+      runCommand("fetch origin " + branch);
```
oferb: What if `fetch origin` fails?

common/repo/GitRepo.java
val-fed: It needs for `commit` command.
This command in our case has next syntax:
`commit -m "commit message"`
If we will split the command just using one space delimiter( command.split(" ") ) we get next parts:
[0] commit
[1] -m
[2] "commit
[3] message"

This code checks the command. If the command contains `"` character it groups all words between `"...."` characters in one part:
[0] commit
[1] -m
[2] commit message

common/repo/GitRepo.java
val-fed: I think we can throw `IllegasStateException` if `fetch origin` fails.
```
      CommandResult fetchCommandResult = runCommand("fetch origin " + branch);
      if (fetchCommandResult.stderr.length() != 0) {
        throw new IllegalStateException(
            "Failed to merge remote branch \'" + branch + "\': " + fetchCommandResult.stderr);
      }
```
What do you think about it?

common/repo/GitRepo.java
vmax: @oferb `mixed` is the default mode (*). we could state it explicitly if you so wish but it won't change the logic

>(*) If you specify the --mixed option, reset will stop at this point. This is also the default, so if you specify no option at all (just git reset HEAD~ in this case), this is where the command will stop.
>
>from https://git-scm.com/book/en/v2/Git-Tools-Reset-Demystified

oferb: Ok, great :)
No need to state it then.

common/repo/GitRepo.java
oferb: We start off with `"commit message"` in a separate string, then we combine it, then we work hard to split it again.

Instead, let's have another method `runCommand()` that gets a `List<String> command` that's already split, and for `commit` command, call it?
The current `runCommand(String command)` would trivially split based on spaces and call `runCommand(List<String> command)`

common/repo/GitRepo.java
oferb: Sounds good

common/repo/GitRepo.java
```diff
@@ -16,65 +16,36 @@
...
-import org.eclipse.jgit.storage.file.FileRepositoryBuilder;
-import org.eclipse.jgit.treewalk.TreeWalk;
 
 // TODO: Implement methods
 @AutoFactory
 public class GitRepo implements Repo {
-  private final Git jGit;
-  private final Repository jGitRepo;
+  private static final String TWO_WHITE_SPACES = "\\s{2}";
+  private static final String DOUBLE_QUOTE = "\"";
```
oferb: It's just a simple string and used only once, so I wouldn't make a `static final` from it.

common/repo/GitRepo.java
```diff
@@ -150,11 +122,12 @@ public void tagHead(String name) {
   }
 
   public ImmutableList<String> getCommitIds(String branch) {
-    CommandResult commandResult = runCommand("log --pretty=%H master~1.." + branch);
+    CommandResult commandResult = runCommand("log --pretty=%H " + branch);
```
oferb: You don't need `master..`?

common/repo/GitRepo.java
```diff
@@ -169,42 +142,25 @@ public void tagHead(String name) {
...
-              untracked ->
-                  files.add(
-                      File.newBuilder().setAction(File.Action.ADD).setFilename(untracked).build()));
-    } catch (GitAPIException e) {
-      throw new RuntimeException(e);
+    CommandResult commandResult = runCommand("status --short");
+    ImmutableList<String> lines = splitLines(commandResult.stdout);
+    for (String line : lines) {
+      String[] parts;
+      if (!line.trim().startsWith("M")
```
oferb: Why only in these cases do you need to remove a space, and why do you later split by spaces?
Can you document by giving an example of the original line?

common/repo/GitRepo.java
```diff
@@ -169,42 +142,25 @@ public void tagHead(String name) {
...
-                          .build()));
-      status
-          .getUntracked()
-          .forEach(
-              untracked ->
-                  files.add(
-                      File.newBuilder().setAction(File.Action.ADD).setFilename(untracked).build()));
-    } catch (GitAPIException e) {
-      throw new RuntimeException(e);
+    CommandResult commandResult = runCommand("status --short");
```
oferb: Did you check `git status --porcelain`?

common/repo/GitRepo.java
```diff
@@ -366,33 +299,65 @@ public void removeBranch(String branch) {
...
-        }
-        // Index 0 should have file data
-        data = objectReader.open(treeWalk.getObjectId(0)).getBytes();
-      }
-      return new String(data, UTF_8);
-    } catch (Exception e) {
-      throw new RuntimeException(e);
+    CommandResult commandResult = runCommand("show " + commitId + ":" + path);
+    String result = commandResult.stdout;
+    int lastIndexOfNewLineSymbol = result.lastIndexOf("\n");
```
oferb: Can you explain this code? (lines 306-311)

common/repo/GitRepo.java
```diff
@@ -366,33 +299,65 @@ public void removeBranch(String branch) {
...
+  }
+
+  public String getHeadCommitId() {
+    CommandResult commandResult = runCommand("rev-parse HEAD");
+    return commandResult.stdout.replace("\n", "");
+  }
+
+  private void switchToMasterBranch() {
+    if (!currentBranch().equals("master")) {
+      System.out.println("You are not in the branch master");
```
oferb: No need to print these. It's normal not to be in the master branch sometimes, and the calling code can decide to log.

common/repo/GitRepo.java
```diff
@@ -366,33 +299,65 @@ public void removeBranch(String branch) {
...
+      switchBranch("master");
+    }
+  }
+
+  @VisibleForTesting
+  public ImmutableList<String> getTagList() {
+    ImmutableList.Builder<String> result = ImmutableList.builder();
+    CommandResult commandResult = runCommand("tag");
+    try {
+      ImmutableList<String> lines = splitLines(commandResult.stdout);
```
oferb: Why are you looping over the lines, to add them to another list, instead of using that other list (`lines`)?

common/repo/GitRepo.java
val-fed: I changed command to `git status --porcelain`.

common/repo/GitRepo.java
val-fed: I need `master..`. Changed it.

common/repo/GitRepo.java
```diff
@@ -16,65 +16,36 @@
...
-import org.eclipse.jgit.revwalk.RevWalk;
-import org.eclipse.jgit.storage.file.FileRepositoryBuilder;
-import org.eclipse.jgit.treewalk.TreeWalk;
 
 // TODO: Implement methods
 @AutoFactory
 public class GitRepo implements Repo {
-  private final Git jGit;
-  private final Repository jGitRepo;
+  private static final String TWO_WHITE_SPACES = "\\s{2}";
```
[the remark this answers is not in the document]
val-fed: `git status --porcelain` (or `git status --short`) command shows lines in a different format. Some lines have two spaces between status symbol and file name, others single spaces. For example:
```
git status --porcelain

D  LICENSE
 M common/repo/GitRepo.java
A  common/repo/new_file_ADD.txt
R  README.md -> rename_README.md
?? common/repo/new_file_UNTRACKED.txt

```

common/repo/GitRepo.java
val-fed: Actually it needed for tests. I some changed tests and deleted this code.

common/repo/GitRepo.java
```diff
@@ -177,18 +149,12 @@ public void tagHead(String name) {
...
     ImmutableList<String> lines = splitLines(commandResult.stdout);
     for (String line : lines) {
-      String[] parts;
-      if (!line.trim().startsWith("M")
-          || !line.trim().startsWith("RM")
-          || !line.trim().startsWith("??")) {
-        line = line.replaceFirst(TWO_WHITE_SPACES, " ");
-      }
-      parts = line.trim().split(" ");
+      line = line.replaceFirst("\\s{2}", " ");
```
oferb: Would `"  "` work instead of `"\\s{2}"`? It's simpler.

common/repo/GitRepo.java
```diff
@@ -177,18 +149,12 @@ public void tagHead(String name) {
...
+      String[] parts = line.trim().split(" ");
 
       File.Builder fileBuilder = File.newBuilder();
```
oferb: I think this is nicer:
```
Action action = getAction(parts[0]);
String filename = action.equals(File.Action.RENAME) ? parts[3] : parts[1]);
files.add(File.newBuilder().setAction(action).setFilename(filename).build());
```

common/repo/GitRepo.java
```diff
@@ -244,107 +211,77 @@ public void tagHead(String name) {
...
   }
 
+  @Override
   public void pushAll() {
-    try {
-      jGit.push().setPushAll().setAtomic(true).call();
-    } catch (GitAPIException e) {
-      throw new RuntimeException(e);
-    }
+    runCommand("push --all origin");
```
oferb: @vmax is this correct?

common/repo/GitRepo.java
```diff
@@ -244,107 +211,77 @@ public void tagHead(String name) {
...
-        add.addFilepattern(file);
-      }
-      try {
-        add.call();
-        jGit.commit().setAll(true).setMessage("Updated changes").call();
-        return true;
-      } catch (NoFilepatternException e) {
-        // Nothing to commit
-        return true;
+    switchToMasterBranch();
```
oferb: @vmax can you review `mergeTheirs()` and `merge()`?

common/repo/GitRepo.java
```diff
@@ -365,23 +302,47 @@ public void removeBranch(String branch) {
...
-    String result = commandResult.stdout;
-    int lastIndexOfNewLineSymbol = result.lastIndexOf("\n");
-    if (lastIndexOfNewLineSymbol >= 0) {
-      result =
-          new StringBuilder(result)
-              .replace(lastIndexOfNewLineSymbol, lastIndexOfNewLineSymbol + 1, "")
-              .toString();
-    }
-    return result;
+    return commandResult.stdout;
```
oferb: You can simplify to:
`return runCommand("show " + commitId + ":" + path).stdout;`

common/repo/GitRepo.java
```diff
@@ -365,23 +302,47 @@ public void removeBranch(String branch) {
...
 
-  public String currentBranch() {
-    return runCommand("rev-parse --abbrev-ref HEAD").stdout.trim();
+  public void addFile(String path) {
+    runCommand("add " + path);
+  }
+
+  public String getHeadCommitId() {
+    CommandResult commandResult = runCommand("rev-parse HEAD");
+    return commandResult.stdout.replace("\n", "");
```
oferb: Maybe also add `trim()`?

common/repo/GitRepo.java
```diff
@@ -365,23 +302,47 @@ public void removeBranch(String branch) {
...
+      switchBranch("master");
+    }
+  }
+
+  @VisibleForTesting
+  public ImmutableList<String> getTagList() {
+    CommandResult commandResult = runCommand("tag");
+    try {
+      return splitLines(commandResult.stdout);
+    } catch (IllegalStateException e) {
```
oferb: The stack trace will include `getTagList()`, so you don't have to explicitly catch & rethrow, plus you're removing the original exception.

common/repo/tests/GitRepoTest.java
```diff
@@ -156,5 +161,149 @@ public void testGetMultipleCommits() {
                 .build()),
         repo.getCommits(TEST_BRANCH));
   }
+
+  @Test
+  public void testAddFile() {
```
oferb: Looks like it's not just `add`, so how about:
s/testAddFile/testCommitFile or testAddAndCommitFile`

common/repo/tests/GitRepoTest.java
```diff
@@ -156,5 +161,149 @@ public void testGetMultipleCommits() {
...
+        repo.getUncommittedFiles());
+  }
+
+  @Test
+  public void testGetUncommittedFilesWhenModifiedFile() {
+    repo.switchBranch(TEST_BRANCH);
+    fileUtils.writeStringUnchecked(TEST_FILE_CONTENTS, fileUtils.joinPaths(repoFolder, TEST_FILE));
+    gitRepo.addFile(TEST_FILE);
+    repo.commit(repo.getUncommittedFiles(), COMMIT_MESSAGE);
+    fileUtils.writeStringUnchecked("new file contains", fileUtils.joinPaths(repoFolder, TEST_FILE));
```
oferb: You probably meant:
`s/new file contains/new file contents`

common/repo/tests/GitRepoTest.java
```diff
@@ -156,5 +161,149 @@ public void testGetMultipleCommits() {
...
+    repo.commit(repo.getUncommittedFiles(), COMMIT_MESSAGE);
+    fileUtils.deleteFileOrDirectoryIfExistsUnchecked(fileUtils.joinPaths(repoFolder, TEST_FILE));
+    assertEquals(
+        ImmutableList.of(
+            File.newBuilder().setFilename(TEST_FILE).setAction(File.Action.DELETE).build()),
+        repo.getUncommittedFiles());
+  }
+
+  @Test
+  public void testTadHead() {
```
oferb: s/testTadHead/testTagHead

common/repo/tests/GitRepoTest.java
```diff
@@ -156,5 +161,149 @@ public void testGetMultipleCommits() {
...
+    repo.tagHead("test_tag");
+    assertEquals(ImmutableList.of("test_tag"), gitRepo.getTagList());
```
oferb: Let's also test 2 tags

common/repo/GitRepo.java
```diff
@@ -153,14 +154,11 @@ public void tagHead(String name) {
...
+      line = line.replaceFirst("  ", " ");
       String[] parts = line.trim().split(" ");
-
-      File.Builder fileBuilder = File.newBuilder();
-      fileBuilder.setAction(getAction(parts[0]));
-      fileBuilder.setFilename(
-          fileBuilder.getAction().equals(File.Action.RENAME) ? parts[3] : parts[1]);
-      files.add(fileBuilder.build());
+      Action action = getAction(parts[0]);
+      String filename = action.equals(File.Action.RENAME) ? parts[3] : parts[1];
```
oferb: Can you add an example of what a regular line and a rename line looks like?
```
// We extract the action and filename from line. Here are some example lines:
// <regular line>
// <rename line>
```

common/repo/GitRepo.java
```diff
@@ -244,107 +211,77 @@ public void tagHead(String name) {
...
+              "merge origin/"
+                  + branch
+                  + " -m Merge_remote-tracking_branch_\'origin/"
+                  + branch
+                  + "\'");
+    } else {
+      mergeCommandResult = runCommand("merge " + branch);
     }
+    String err = mergeCommandResult.stderr;
+    return err.length() == 0;
```
vmax: @val-fed `return mergeCommandResult.stderr.length().isEmpty()`

common/repo/GitRepo.java
```diff
@@ -244,107 +211,77 @@ public void tagHead(String name) {
...
+    String err = commandResult.stderr;
+    return err.length() == 0;
+  }
+
+  public boolean merge(String branch, boolean remote) {
+    switchToMasterBranch();
+    CommandResult mergeCommandResult;
+    if (remote) {
+      CommandResult fetchCommandResult = runCommand("fetch origin " + branch);
+      if (fetchCommandResult.stderr.length() != 0) {
```
vmax: @val-fed `if (!fetchCommandResult.stderr.isEmpty())`

common/repo/GitRepo.java
vmax: @val-fed @oferb  did you decide to not reset `HEAD` to previous position?

common/repo/GitRepo.java
oferb: No we didn't. Good catch!

common/repo/repo.proto
val-fed: Done.

common/repo/GitRepo.java
val-fed: Done.

common/repo/GitRepo.java
val-fed: Done.

common/repo/GitRepo.java
val-fed: Done.

common/repo/GitRepo.java
val-fed: Done.

common/repo/GitRepo.java
val-fed: Changed.

common/repo/GitRepo.java
val-fed: Fixed.

common/repo/GitRepo.java
val-fed: Fixed.

common/repo/GitRepo.java
val-fed: Fixed.

common/repo/tests/GitRepoTest.java
val-fed: Fixed.

common/repo/tests/GitRepoTest.java
val-fed: Fixed.

common/repo/tests/GitRepoTest.java
val-fed: Fixed.

common/repo/tests/GitRepoTest.java
val-fed: Done.

common/repo/GitRepo.java
val-fed: Done.

common/repo/GitRepo.java
val-fed: Fixed.

PR #181, opened 2018-07-19
Scheduled pull on all head repos

vmax:


tools/local_server/LocalServer.java
```diff
@@ -46,8 +51,46 @@
...
 
+  public static class HeadUpdater extends TimerTask {
+
+    private final FileUtils fileUtils;
+    private final Config config;
+    private GitRepoFactory repoFactory;
+    private String currentWorkspaceName;
+
+    @Inject
+    public HeadUpdater(FileUtils utils, Config config, GitRepoFactory repoFactory) {
```
oferb: You can use `@Named("Base path")`

Issue #167, opened 2018-07-11
Move completion.sh to tools


oferb:
This one is really trivial - 
Move
https://github.com/google/startup-os/blob/master/completion.sh
To `tools`

NikleshPhabiani:
Done.

oferb:
:)

PR #183, opened 2018-07-19
Move completion.sh to tools - #167

NikleshPhabiani:
Moved completion.sh to **tools**


tools/completion.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
oferb: Thanks!
Did you delete it from the repo root?
Also, please update the "link" here:
https://github.com/google/startup-os/blob/master/commands.txt#L41

oferb:
Just one last small thing - 
Also, please update the "link" here to the new location:
https://github.com/google/startup-os/blob/master/commands.txt#L41

Thanks! :-)

tools/completion.sh
NikleshPhabiani: I had accidentally committed from **tools** and hence it missed the deletion. I have incorporated it now. I even did some experimentation with merging into my branch since I was not able to push my changes because of another commit as a part of this pull request. Hope that does not impact anything. Kindly let me know.

oferb:
Thanks a lot! Merging now :)

NikleshPhabiani:
Pleasure :)

PR #186, opened 2018-07-20
Fix readFile method in FileUtils

val-fed:
Fix bug with removing last `\n`


common/FileUtils.java
```diff
@@ -185,7 +185,14 @@ public String getCurrentWorkingDirectoryName() {
...
   /** Reads a text file. */
   public String readFile(String path) throws IOException {
-    return String.join("\n", Files.readAllLines(fileSystem.getPath(expandHomeDirectory(path))));
+    ImmutableList<String> lines =
+        ImmutableList.copyOf(Files.readAllLines(fileSystem.getPath(expandHomeDirectory(path))));
+    if (lines.isEmpty()) {
+      return "";
+    } else {
+      String lastLine = lines.get(lines.size() - 1);
+      return lastLine.equals("\n") ? String.join("\n", lines) + "\n" : String.join("\n", lines);
```
oferb: Please add a TODO:
`// TODO: Add supoprt for '/r/n' for Windows.`

val-fed: I changed `\n` to `System.lineSeparator()`. It should resolve differences between OS.

PR #190, opened 2018-07-21
Fix methods in FileUtils

val-fed:
- fix `writePrototxt`
- fix `writeString`
- fix `readFile`
- delete `tempFix` in CodeReviewServiceTest
- add more tests to `FileUtilsTest`


common/FileUtils.java
```diff
@@ -185,7 +204,16 @@ public String getCurrentWorkingDirectoryName() {
...
   public String readFile(String path) throws IOException {
-    return String.join("\n", Files.readAllLines(fileSystem.getPath(expandHomeDirectory(path))));
+    ImmutableList<String> lines =
+        ImmutableList.copyOf(Files.readAllLines(fileSystem.getPath(expandHomeDirectory(path))));
+    if (lines.isEmpty()) {
+      return "";
+    } else {
+      String lastLine = lines.get(lines.size() - 1);
+      return lastLine.equals(System.lineSeparator())
+          ? String.join(System.lineSeparator(), lines) + System.lineSeparator()
```
oferb: I think we should revert back to `/n`. The content of the file won't change based on OS, so if it's right on Windows, it'll be wrong on Linux and vice versa.
Let's stick to `\n` and add a TODO to look into it later.

PR #191, opened 2018-07-22
Use diff-patch-match for TextDifferencer

oferb:


common/TextDifferencer.java
```diff
@@ -29,7 +29,6 @@
 import javax.swing.text.Segment;
 import java.util.LinkedList;
```
val-fed: Unused imports:
- import java.util.ArrayList;
- import java.util.Collections;
- import java.util.List;
- import java.util.stream.Collectors;
- import java.util.stream.IntStream;
- import javax.swing.text.Segment;

common/tests/TextDifferencerTool.java
```diff
@@ -32,7 +32,6 @@
 import name.fraser.neil.plaintext.DiffMatchPatch.Operation;
 import java.util.LinkedList;
```
val-fed: Unused imports:
- import com.google.startupos.common.TextChange.Type;
- import com.google.startupos.name.fraser.neil.plaintext.DiffMatchPatch;
- import com.google.startupos.name.fraser.neil.plaintext.DiffMatchPatch.Operation;
- import java.util.LinkedList;

common/DiffMatchPatch.java
```diff
@@ -31,91 +31,74 @@
...
  * Source: https://github.com/google/diff-match-patch
  * TODO: Add diff-patch-match to WORKSPACE and use directly instead of copying the code.
  */
 
-/**
- * Class containing the diff, match and patch methods.
- * Also contains the behaviour settings.
- */
+/** Class containing the diff, match and patch methods. Also contains the behaviour settings. */
 public class DiffMatchPatch {
```
val-fed: Looks like package name is incorrect. Is it ok?
Maybe the name should be `com.google.startupos.common`?

val-fed: Maybe replace `import java.util.*` to single class import?

common/DiffMatchPatch.java
```diff
@@ -146,19 +128,19 @@ protected LinesToCharsResult(String chars1, String chars2,
...
+   * @param checklines Speedup flag. If false, then don't run a line-level diff first to identify
+   *     the changed areas. If true, then run a faster slightly less optimal diff.
+   * @param deadline Time when the diff should be complete by. Used internally for recursive calls.
+   *     Users should set DiffTimeout instead.
    * @return Linked List of Diff objects.
    */
-  private LinkedList<Diff> diff_main(String text1, String text2,
-                                     boolean checklines, long deadline) {
+  private LinkedList<Diff> diff_main(
+      String text1, String text2, boolean checklines, long deadline) {
```
val-fed: The explicit type argument `<Diff>` (`diffs = new LinkedList<Diff>()`) can be replaced with `<>` (line 152). 
Also below.

common/TextDifferencer.java
oferb: Done

common/tests/TextDifferencerTool.java
oferb: Done

common/DiffMatchPatch.java
oferb: As noted, this code is taken as-is from another repo.

PR #200, opened 2018-07-25
ReviewCommand should perform pushAll on all repos

vmax:


tools/aa/commands/ReviewCommand.java
```diff
@@ -16,24 +16,42 @@
...
+  private String workspacePath;
+
   private final CodeReviewServiceGrpc.CodeReviewServiceBlockingStub codeReviewBlockingStub;
   private final Integer currentDiffNumber;
 
   @Inject
-  public ReviewCommand(@Named("Current diff number") Integer currentDiffNumber) {
+  public ReviewCommand(
+      FileUtils utils,
+      Protos.Config config,
```
oferb: Please import `Protos.Config` and use `Config` here.

PR #201, opened 2018-07-25
Remove `npm install`

vmax:
@oferb I bumped up cache version so that I'm sure `npm install` done previously wouldn't affect this particular build (not likely to happen) & to cache new repos we are using to not re-download them (`@startupos_binaries`, `@protoc`…)

### DIFF
--- a/commands.txt
+++ b/commands.txt
@@ -1,5 +1,4 @@
 # Build every buildable Bazel target and test every testable target
-# TODO: Until we move to Angular Bazel, execute `npm install` beforehand
 bazel query //... | grep -v ^\/\/third_party | grep -v *node_modules | xargs bazel test
 
 # Format dependencies.yaml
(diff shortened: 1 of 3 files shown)


commands.txt
```diff
@@ -1,6 +1,8 @@
-# Build every buildable Bazel target and test every testable target
-# TODO: Until we move to Angular Bazel, execute `npm install` beforehand
-bazel query //... | grep -v ^\/\/third_party | grep -v *node_modules | xargs bazel test
+# Build every buildable Bazel target
```
oferb: I think we still need to ignore `npm install` because to run Angular server, it's still needed, and so will still create the same `node_modules` directory. What we don't need anymore is the `npm install` before we run `bazel build`. So I think the only change here is removing the TODO.

PR #204, opened 2018-07-26
Add shutdown logic to aa tool

val-fed:


tools/aa/commands/InitCommand.java
```diff
@@ -86,11 +97,42 @@ public boolean run(String[] args) {
...
+      revertChanges();
       return false;
     } catch (Exception e) {
+      revertChanges();
       e.printStackTrace();
     }
     return true;
   }
+
+  private void revertChanges() {
```
oferb: I think in this case, init exits when `basePath.get()` contains something, so all we need to check is whether it existed.
If it did, leave it empty. If it didn't, delete it and all of its contents.

tools/aa/commands/InitCommand.java
```diff
@@ -56,17 +62,17 @@ public InitCommand(FileUtils fileUtils, GitRepoFactory gitRepoFactory) {
...
         System.out.println("Error: Base folder exists and is not empty");
         System.exit(1);
       }
+      wasBaseFolderExisted = fileUtils.folderExists(basePath.get());
       // Create folders
-      fileUtils.mkdirs(fileUtils.joinPaths(basePath.get(), "head"));
-      fileUtils.mkdirs(fileUtils.joinPaths(basePath.get(), "ws"));
-      fileUtils.mkdirs(fileUtils.joinPaths(basePath.get(), "local"));
-      fileUtils.mkdirs(fileUtils.joinPaths(basePath.get(), "logs"));
+      fileUtils.mkdirs(fileUtils.joinPaths(basePath.get(), HEAD_FOLDERNAME));
```
oferb: For strings used only once, it's mostly ok to not create a `static final`, since it's easier to read it when the string is there.

tools/aa/commands/InitCommand.java
```diff
@@ -56,17 +62,17 @@ public InitCommand(FileUtils fileUtils, GitRepoFactory gitRepoFactory) {
   public boolean run(String[] args) {
     // TODO: Add Flags.parse() support for specifying a particular class, not a whole package
     Flags.parse(args, InitCommand.class.getPackage());
-
     try {
       if (!fileUtils.folderEmptyOrNotExists(basePath.get())) {
         System.out.println("Error: Base folder exists and is not empty");
         System.exit(1);
       }
+      wasBaseFolderExisted = fileUtils.folderExists(basePath.get());
```
oferb: s/wasBaseFolderExisted/baseFolderExistedBefore

tools/aa/commands/SyncCommand.java
```diff
@@ -105,9 +110,18 @@ public boolean run(String[] args) {
...
+      revertChanges(workspacePath, wasTempBranchForSyncExisted);
       e.printStackTrace();
     }
     return true;
   }
+
+  private void revertChanges(String workspacePath, boolean wasTempBranchForSyncExist) {
+    GitRepo repo = repoFactory.create(workspacePath);
+    ImmutableList<String> branches = repo.listBranches();
+    if (branches.contains(TEMP_BRANCH_FOR_SYNC) && !wasTempBranchForSyncExist) {
```
oferb: Actually in this case, let's leave the temp branch, because we're not really undoing everything, and if there was an error, the temp branch may still have information that is otherwise lost.

We should, however, switch to the original branch.

tools/aa/commands/tests/WorkspaceCommandTest.java
```diff
@@ -137,12 +137,18 @@ public void forcedWorkspaceCommandTest() throws Exception {
...
     String[] args = {"workspace", "-f", "workspace_name"};
     workspaceCommand.run(args);
-    assertTrue(fileUtils.folderExists("/base/ws/workspace_name"));
+
     // For some reason, in the tests, fileUtils.copyDirectoryToDirectory() throws:
     // NoSuchFileException: /base/head
     // It might be a bug in jimfs, since it works in practice.
     // For this reason, we're not checking the copying of the repos from head to workspace and cd:
     // assertEquals("cd /base/ws/workspace_name\n", outContent.toString());
+
```
oferb: Why is there an issue currently?

val-fed:
```
private boolean createWorkspace() {
    if (fileUtils.folderExists(workspacePath)) {
      System.err.println(RED_ERROR + "Workspace already exists");
      return false;
    } else {
      fileUtils.mkdirs(workspacePath);
      try {
        fileUtils.copyDirectoryToDirectory(
            fileUtils.joinPaths(config.getBasePath(), "head"), workspacePath, "^bazel-.*$");
      } catch (IOException e) {
        fileUtils.deleteDirectoryUnchecked(workspacePath);
        e.printStackTrace();
        return false;
      }
    }
    return true;
  }
```
`createWorkspace()` method contains `copyDirectoryToDirectory` method which throws `NoSuchFileException` in `jimfs`. If `IOException` is thrown workspace directory is deleted. And test failed.

tools/aa/commands/SyncCommand.java
val-fed: Done.

tools/aa/commands/InitCommand.java
val-fed: Done.

tools/aa/commands/InitCommand.java
val-fed: Done.

tools/aa/commands/InitCommand.java
val-fed: Done.

tools/aa/commands/InitCommand.java
```diff
@@ -45,6 +45,7 @@
 
   private final GitRepoFactory gitRepoFactory;
   private FileUtils fileUtils;
+  private boolean baseFolderExistedBefore = true;
```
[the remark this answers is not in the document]
val-fed: If during getting `baseFolderExistedBefore = fileUtils.folderExists(basePath.get())`(line 65) happens Exception and `baseFolder` existed before, it guarantees that `baseFolder` won't be deleted.

PR #205, opened 2018-07-27
Refactor and fix static analysis issues

oferb:


tools/aa/commands/DiffCommand.java
```diff
@@ -42,9 +41,9 @@
 public class DiffCommand implements AaCommand {
   private final FileUtils fileUtils;
   private final GitRepoFactory gitRepoFactory;
-  private String currentWorkspaceName;
+  private String workspaceName;
   private String workspacePath;
-  private Integer currentDiffNumber;
+  private Integer diffNumber;
 
   private static final Integer GRPC_PORT = 8001;
```
val-fed: Do we need `public` access to flags?

oferb: Good question :)

PR #208, opened 2018-07-28
Fix GetDocument to read protobin format, fail gracefully if branch not in head

oferb:


common/firestore/FirestoreClient.java
```diff
@@ -60,30 +60,47 @@ private String getCreateDocumentUrl(String user) {
...
+    connection.setRequestMethod("GET");
+    connection.setRequestProperty("Authorization", "Bearer " + token);
+    try (BufferedReader reader =
+        new BufferedReader(new InputStreamReader(connection.getInputStream()))) {
+      String line;
+      while ((line = reader.readLine()) != null) {
+        response.append(line);
+      }
+    }
+    if (connection.getResponseCode() != HTTP_OK) {
```
vmax: @oferb shouldn't you check response code _before_ trying to read from it?

oferb: The read `getInputStream()` actually does the call. Kind of odd but that's how it works.

https://stackoverflow.com/questions/16122999/java-urlconnection-when-do-i-need-to-use-the-connect-method

In the case of an error, it'll read the contents into `response` and then fail. I think it's not so bad to leave as is.

PR #209, opened 2018-07-29
Various checks for dependencies.yaml

vmax:


check_dependencies.sh
```diff
@@ -0,0 +1,50 @@
...
+
+bazel run //tools/formatter -- \
+  --path $(pwd)/third_party/maven/ \
+  --build \
+  &>/dev/null
+
+# Prints out an error only if both conditions are satisfied:
+# * we are on CircleCI
+# * working tree contains unstaged changes
+# When ran locally it silently fixes everything.
```
oferb: It always fixes everything, just when ran locally it does it silently, right?

vmax: @oferb this is correct

check_dependencies.sh
```diff
@@ -0,0 +1,48 @@
...
+
+BAZEL_DEPS_CACHE="$HOME/bazel_deps_cache"
+DEP_CHECKSUM=($(sha1sum $(pwd)/dependencies.yaml))
+CACHED_RESULT="$BAZEL_DEPS_CACHE/$DEP_CHECKSUM"
+
+if [[ ! -z "$CIRCLECI" ]]; then
+  mkdir -p $BAZEL_DEPS_CACHE
+  isok=$(<$CACHED_RESULT)
+
+  if [ "$isok" == "OK" ]; then
```
oferb: This cache mechanism allows to "come back" to previously cached `dependencies.yaml`, even if they don't match the current `third_party/maven`.
I think what we need is to somehow figure out if `dependencies.yaml` was changed in the current PR and if so, rerun this. Not sure how to find that out though..

Alternatively, we could either:
* Run it always
* Check the whitelist for dependencies in https://github.com/google/startup-os/blob/master/third_party/maven/workspace.bzl. This doesn't check the correspondence, but at least checks the whitelist.

vmax: @oferb 
>figure out if `dependencies.yaml` was changed in the current PR and if so, rerun this

maybe `git diff` to `master` may help

oferb: Actually, looking at `workspace.bzl` I think it'll be pretty easy to do a whitelist check on that.
If `git diff` on `master` works then great, but looking at `workspace.bzl` I think we should do both. They do different things.

tools/fix_formatting.sh
```diff
@@ -7,11 +7,15 @@
 RED=$(tput setaf 1)
 RESET=$(tput sgr0)
 
+bazel run @bazel_deps//:parse -- format-deps \
```
oferb: I guess we can use `@startupos-binaries` and remove `@bazel_deps`.

tools/deps/check_dependencies.sh
```diff
@@ -0,0 +1,37 @@
...
+RED=$(tput setaf 1)
+GREEN=$(tput setaf 2)
+RESET=$(tput sgr0)
+
+git diff --name-only origin/master | grep --quiet dependencies.yaml
+if [ $? -eq 0 ]
+then
+  echo "$RED[!] ""dependencies.yaml was modified, running the check$RESET"
+else
+  echo "$GREEN[!] ""dependencies.yaml was not modified, exittin (exit code $?) $RESET"
```
oferb: s/exittin/exiting

tools/deps/check_dependencies.sh
```diff
@@ -0,0 +1,37 @@
...
+  -d dependencies.yaml \
+  &>/dev/null
+
+# Format generated BUILD files
+bazel run //tools/formatter -- \
+  --path $(pwd)/third_party/maven/ \
+  --build \
+  &>/dev/null
+
+# Print error if on CircleCI
```
oferb: Print error if on CircleCI and dependencies were not up-to-date.

PR #210, opened 2018-07-31
Read and write bytes in FileUtils

oferb:


common/tests/FileUtilsTest.java
```diff
@@ -543,15 +542,14 @@ public void testReadPrototxt() throws IOException {
...
-            "}",
-            "enum_field: YES"),
-        UTF_8);
+        ("int32_field: 123\n"
+                + "string_field: \"foo\"\n"
+                + "map_field {\n"
+                + "key: \"foo\"\n"
+                + "value: 123\n"
+                + "}\n"
+                + "enum_field: YES")
```
val-fed: Data which is written to the prototxt file in `testReadPrototxt()` method:
```
Files.write(
        testPath,
        ("int32_field: 123\n"
                + "string_field: \"foo\"\n"
                + "map_field {\n"
                + "key: \"foo\"\n"
                + "value: 123\n"
                + "}\n"
                + "enum_field: YES")
            .getBytes(UTF_8));
```
and `expected` result in `testWritePrototxt()` method:

```
String expected =
        "int32_field: 123\n"
            + "string_field: \"foo\"\n"
            + "map_field {\n"
            + "  key: \"foo\"\n"
            + "  value: 123\n"
            + "}\n"
            + "enum_field: YES\n";
```
are identical. Difference only in last `\n` symbol. I think will be better if `\n` symbol will be added to data in `testReadPrototxt()` method.

oferb: I think it's easier to understand each test as it is now. If we introduce a common  `static final`, you won't see the string and the rest of the code on the screen at the same time.

common/tests/FileUtilsTest.java
```diff
@@ -107,7 +107,7 @@ public void testFolderExistsIsFalseWhenFile() throws IOException {
     }
```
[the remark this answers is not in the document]
oferb: Fixed: https://github.com/google/startup-os/pull/215

oferb:
Closing this as it's been added to a different PR

PR #213, opened 2018-08-01
Remove C++ code from repo

vmax:


compile.sh
```diff
@@ -9,20 +9,5 @@ if [[ $1 != "build" && $1 != "test" ]]; then
...
-# already-built @com_google_protobuf//:protobuf
-
-if [[ -z "$CIRCLECI" ]]; then
-  DELETED_PACKAGES="";
-else
-  DELETED_PACKAGES="--deleted_packages $(cat .circleci/ignored_bazel_packages.txt)";
-fi
-
-bazel $1 $DELETED_PACKAGES //...
+bazel $1 //...
```
oferb: Do we still need this file then?
We can just do
`bazel build //..`
`bazel test //..`

vmax: @oferb we do for two reasons:
* we'll also check for Android SDK [in other PR](https://github.com/google/startup-os/blob/584c3fb5d81ff3f4b57dfa8050fddaeaec42f513/compile.sh#L12-L16)
* it'll be easier to customize building process if we ever need to

oferb: Ok

get-android-sdk.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
oferb: Why is this one here?

PR #215, opened 2018-08-01
Remove unused import

oferb:


jbduncan:
@oferb If you're using the google-java-format command-line application to format Java code, you can use `--dry-run` and `--set-exit-if-changed` as described in https://github.com/google/google-java-format/pull/106, and then check if the exit code is 0 (code is already formatted) or 1 (code is _unformatted_).

jbduncan:
Sorry, not @oferb, I meant @vmax. :)

PR #216, opened 2018-08-02
Reviewer - diff page redesign

webdevelopland:
What's new:
- design of diff page is rebuild
- method of adding/removing users from reviewer/cc list is changed.
Now it's based on modal UI.
- name and location of some components were changed.
- major changes in the code. A lot of features implemented in a different way.
- standardization of design through a whole app
- new icon of web app :)


tools/reviewer/web/README.md
```diff
@@ -2,7 +2,8 @@
 Run `ng install` for the first time.
 
 Run `npm run protoc` to generate proto functions.
-You need installed google-protobuf proto3 on your computer.
+You need installed google-protobuf version 3.5 or above on your computer.
```
oferb: Let's rephrase:
You need google-protobuf 3.5 or above installed on your machine.
Supported on Linux and Mac only.

tools/reviewer/web/README.md
```diff
@@ -2,7 +2,8 @@
...
+Supported on Linux and Mac machines only.
```
oferb: How about we mention Linux & Mac support  at the bottom?

tools/reviewer/web/src/app/dashboard/diff/diff-discussion/diff-discussion.component.ts
```diff
@@ -0,0 +1,38 @@
...
+import { AuthService } from '@/shared/services';
+import { DiffService } from '../diff.service';
+
+@Component({
+  selector: 'diff-discussion',
+  templateUrl: './diff-discussion.component.html',
+  styleUrls: ['./diff-discussion.component.scss'],
+  providers: [DiffService],
+})
+export class DiffDiscussionComponent {
```
oferb: Can you please document what each component is?
One way to do it is to add a screenshot of the webapp with a rectangle on the component (if it's a visible component). We can host the image in some image hosting service of your choice (or Firebase hosting even).

tools/reviewer/web/src/app/dashboard/diff/diff-header/add-user-dialog/add-user-dialog.component.ts
```diff
@@ -0,0 +1,51 @@
...
+    this.dialogRef.close();
+  }
+
+  isEmailValid(email: string): boolean {
+    if (!email) {
+      return false;
+    }
+
+    // Example: my_super-name@flip-flop.com
+    const emailValidation: RegExp = /^[\w-.]+@[\w-]+\.[\w]+$/;
```
oferb: Where is this regex from?
This one seems more complicated: http://emailregex.com/
Are you sure we're covering all cases?

tools/reviewer/web/src/app/dashboard/diff/diff-header/add-user-dialog/add-user-dialog.component.ts
```diff
@@ -0,0 +1,51 @@
...
+    return emailValidation.test(email.toLowerCase());
+  }
+
+  hideError() {
+    this.isEmailInvalid = false;
+  }
+
+  @HostListener('window:keydown', ['$event'])
+  handleKeyDown(event: KeyboardEvent) {
+    // Hide error when user's typing something
```
oferb: s/user's/user is
To avoid confusion with ownership: "user's shoes".

tools/reviewer/web/src/app/dashboard/diff/diff-header/diff-header-content/diff-header-content.component.ts
```diff
@@ -0,0 +1,143 @@
...
+  @Input() diff: Diff;
+
+  constructor(
+    public authService: AuthService,
+    public firebaseService: FirebaseService,
+    public notificationService: NotificationService,
+    public diffHeaderService: DiffHeaderService,
+    public dialog: MatDialog,
+  ) {
+    // Save description when it's changed
```
oferb: This is cool, but I'm not sure it's the desired behavior.
What if the user wants to undo?

tools/reviewer/web/src/app/dashboard/diff/diff-header/diff-header-content/diff-header-content.component.ts
```diff
@@ -0,0 +1,143 @@
...
+      const message: string = reviewer.getNeedsAttention() ?
+        `Attention of ${username} is requested` :
+        `Attention of ${username} is canceled`;
+      this.notificationService.success(message);
+    }, () => {
+      this.notificationService.error("Attention isn't changed");
+    });
+  }
+
+  // Is the email is present in the CC list?
```
oferb: Is the email present in the CC list?

tools/reviewer/web/src/app/dashboard/diff/diff-header/diff-header-content/diff-header-content.component.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  saveUserToFirebase(email: string): void {
+    this.updateUserListInFirebase(email, 'added');
+  }
+
+  removeUserFromFirebase(email: string): void {
+    this.updateUserListInFirebase(email, 'removed');
+  }
+
+  // To make code more DRY
```
oferb: >// To make code more DRY
> // saveUserToFirebase and removeUserFromFirebase use the method

It's good that you did this, but I think there's no need to mention it.

tools/reviewer/web/src/app/app.module.ts
```diff
@@ -1,20 +1,29 @@
...
 import { HttpModule } from '@angular/http';
-import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+import {
+  BrowserAnimationsModule,
+} from '@angular/platform-browser/animations';
 
 import { AngularFireModule } from 'angularfire2';
-import { AngularFireAuth } from 'angularfire2/auth';
 import { AngularFireAuthModule } from 'angularfire2/auth';
+import { AngularFireAuth } from 'angularfire2/auth';
```
MateMahinKhan: something very trivial, both imports are from `angularfire2/auth;`

tools/reviewer/web/src/app/dashboard/diff/diff-header/add-user-dialog/add-user-dialog.component.ts
jbduncan: Is this regex being used just to let the user know if the email address they entered is _likely_ to be valid, or is it being used to _force_ the user the enter a valid address?

If it's the latter, this is bad practice for reasons described in https://stackoverflow.com/a/202528/2252930.

But if it's the former, then please, do carry on! :)

tools/reviewer/web/src/app/dashboard/diff/diff-discussion/diff-discussion.component.html
```diff
@@ -0,0 +1,53 @@
...
+    >Expand</button>
+  </div>
+</div>
+<div class="cr-header">
+  <mat-table [dataSource]="threads">
+      <ng-container matColumnDef="discussions">
+        <mat-header-cell *matHeaderCellDef>
+          General discussions
+          ({{getAmountOfComments()}} comments,
+          {{getAmountOfComments()}} unresolved)
```
MateMahinKhan: same function to get the comments and unresolved comments?

tools/reviewer/web/src/app/dashboard/diff/diff-discussion/diff-discussion.component.html
```diff
@@ -0,0 +1,53 @@
...
+            {{ thread.getFile().getFilename() }}
+            ({{ thread.getCommentList().length }} comments)
+          </div>
+          <div
+            class="comment"
+            *ngFor="let comment of thread.getCommentList()"
+          >
+            <div class="username">{{ comment.getContent() }}</div>
+            <div class="message">
+              by <span class="bold">{{ getUsername(comment) }}</span>
```
MateMahinKhan: The div that has class `username` is displaying the comment contents and the div that has class `message` is displaying the meta data for the comment.

MateMahinKhan: Also, something worth discussing. @oferb should there be classes in HTML that don't have any styles attached to them. They do help in identifying what functionality is encapsulated in the HTML element but they could also create too many class names.

tools/reviewer/web/src/app/dashboard/diff/diff-header/reply-popup/reply-popup.component.html
```diff
@@ -0,0 +1,37 @@
...
+      [(ngModel)]="approved" 
+    >Approved</mat-checkbox>
+    
+    <button
+      class="cr-light-button"
+      mat-raised-button
+      color="accent"
+      (click)="reply()"
+      [disabled]="isLoading"
+    >Reply</button>
```
MateMahinKhan: Wasn't this button `Submit`? did it get changed to `Reply`?

tools/reviewer/web/src/app/dashboard/page-loading/page-loading.component.ts
```diff
@@ -0,0 +1,10 @@
+import { Component, Input } from '@angular/core';
```
MateMahinKhan: I was thinking if page loading doesn't have to do anything with our diffs logic it can be moved to layout folder.

tools/reviewer/web/src/app/dashboard/diff/diff-header/add-user-dialog/add-user-dialog.component.ts
oferb: Force the user to enter a valid email address.
We're not planning to send emails to this address (at least not without validating it as is mentioned in the link), but if we can prevent at least _some_ manual entry mistakes, I think we should.
If it ends up being a problem that people are entering wrong emails, we'll think of a more elaborate solution.

tools/reviewer/web/src/app/dashboard/diff/diff-discussion/diff-discussion.component.html
oferb: Readability of code is very important. What's the drawback in having many class names?

tools/reviewer/web/src/app/dashboard/diff/diff-header/add-user-dialog/add-user-dialog.component.ts
jbduncan: @oferb Sorry, I'm a bit lost.

Are you advocating forcing a user to enter an email address that matches the regex, no exceptions?

Or are you suggesting that, like most other sites (that I've personally used), it should just act as a suggestion for the user, allowing them to enter their email address even if it doesn't match the regex exactly?

tools/reviewer/web/README.md
webdevelopland: Well, most of the app supports windows. Only `npm run protoc` doesn't. For example, if I have generated proto functions, I can continue development on windows without any problems.

tools/reviewer/web/src/app/dashboard/diff/diff-header/add-user-dialog/add-user-dialog.component.ts
webdevelopland: Yeah, I know there're much more difficult email regexes, but to be honest, I made the validation only to be sure that email has basic email structure and we can take username from it.

tools/reviewer/web/src/app/dashboard/diff/diff-header/diff-header-content/diff-header-content.component.ts
webdevelopland: Yeah, you're right?
What is the desired behavior though?

tools/reviewer/web/src/app/dashboard/diff/diff-header/diff-header-content/diff-header-content.component.ts
webdevelopland: :D

tools/reviewer/web/src/app/dashboard/diff/diff-header/diff-header-content/diff-header-content.component.ts
webdevelopland: Other developers could find the method and wonder what is the purpose of it.

tools/reviewer/web/src/app/dashboard/diff/diff-discussion/diff-discussion.component.html
webdevelopland: Do we have a way to know an amount of unresolved *comments*?

tools/reviewer/web/src/app/dashboard/page-loading/page-loading.component.ts
webdevelopland: I tried, but component from one module can't be used in another.

tools/reviewer/web/src/app/dashboard/diff/diff-header/add-user-dialog/add-user-dialog.component.ts
jbduncan: @webdevelopland That sounds fair to me. :+1:

oferb: Why would a user want to enter an email that the regex rejects?
That would imply the regex is not good, and in that case, we should use a different one :)

tools/reviewer/web/src/app/dashboard/diff/diff-header/diff-header-content/diff-header-content.component.ts
oferb: I think the same as other fields - click on edit -> edit -> save/cancel.

tools/reviewer/web/src/app/dashboard/diff/diff-header/diff-header-content/diff-header-content.component.ts
oferb: They can see it's used in the same file, no?

tools/reviewer/web/src/app/dashboard/diff/diff-header/add-user-dialog/add-user-dialog.component.ts
jbduncan: @oferb As you kind of implied, a user would expect a website to allow them to enter their email address even if it deviates from whatever preconceived notion of an "email address" the website may have, but is otherwise a perfectly valid email address. In other words, the way I see it, the website should not use a regex or whatever as the absolute validator; it should instead use the regex to _guide_ the user on the UI side, and then after the user has hit Enter, the website should send an email to that email address so that the user can confirm that they've entered the correct address.

I was burned by something kind of similar in the past, where I wasn't able to use my email address to create an account on a website (which I forget the name of now) because my email address was considered too long, even though it was legitimate. Now, I know, it's not the same problem as having an incorrect regex, but I think it still shows that, ultimately, the final validation _shouldn't_ happen on the website itself, but rather by sending a confirmation email to the entered email address.

tools/reviewer/web/src/app/dashboard/diff/diff-discussion/diff-discussion.component.ts
webdevelopland: I'm not sure about the idea with screenshots. I added them, but I think it's difficult to maintain the screenshots up to date.

tools/reviewer/web/src/app/dashboard/diff/diff-discussion/diff-discussion.component.html
MateMahinKhan: yeah I think there was. Thread has a property isDone. I think I discussed it with Ofer some time back. We should ask @oferb just to be sure.

tools/reviewer/web/src/app/dashboard/diff/diff-discussion/diff-discussion.component.html
MateMahinKhan: having many classes might clutter the code. That's all I could think of.

tools/reviewer/web/src/app/dashboard/page-loading/page-loading.component.ts
MateMahinKhan: we could put it in the shared module... but then again it's just one component. We can put it off until there are more components like it.

tools/reviewer/web/src/app/layout/toolbar/toolbar.component.ts
```diff
@@ -7,7 +7,7 @@ import { Component } from '@angular/core';
   styleUrls: ['./toolbar.component.scss'],
```
MateMahinKhan: We can change the selector here.

tools/reviewer/web/src/app/dashboard/diff/diff-discussion/diff-discussion.component.html
webdevelopland: IsDone is a field of a thread, not comment.
https://github.com/google/startup-os/blob/master/tools/reviewer/service/code_review.proto#L105

tools/reviewer/web/src/app/dashboard/diff/diff-header/add-user-dialog/add-user-dialog.component.ts
oferb: Got it. I didn't know email format was so ill-defined. The link I shared mentions `RFC 5322 Official Standard`. I thought there's an accepted standard for email addresses, but I guess I'm wrong. In that case, let's indeed stay with a minimally-restrictive regex.

tools/reviewer/web/src/app/dashboard/diff/diff-discussion/diff-discussion.component.ts
oferb: It's true. It's just to get a basic understanding of where the component is on the page. Might have been an illustration or sketch, but it's easier to take a screenshot.

tools/reviewer/web/README.md
oferb: Got it.

tools/reviewer/web/src/app/dashboard/diff/diff-discussion/diff-discussion.component.html
oferb: Not sure I follow, but indeed IsDone is a thread field, and so unresolved is a thread-related status. You could say the last comment in an unresolved thread is unresolved.

tools/reviewer/web/src/app/dashboard/diff/diff-discussion/diff-discussion.component.html
oferb: I don't feel strongly about this. The guiding principle is that the code should be fun to work with, for the author and for others.

tools/reviewer/web/src/app/dashboard/diff/diff-header/diff-header-content/diff-header-content.component.ts
```diff
@@ -0,0 +1,148 @@
...
+import {
+  AuthService,
+  FirebaseService,
+  NotificationService,
+} from '@/shared/services';
+import { AddUserDialogComponent } from '../add-user-dialog';
+import { DiffHeaderService } from '../diff-header.service';
+
+// The component implements content of the header
+// How it looks: "/src/assets/design-blocks/header-content.jpg"
```
oferb: Let's not add these images to the repo. They just take up space. How about we host them in an image hosting service? (e.g imgur).

tools/reviewer/web/src/app/dashboard/diff/diff-header/add-user-dialog/add-user-dialog.component.ts
jbduncan: > The link I shared mentions RFC 5322 Official Standard. I thought there's an accepted standard for email addresses.

And I think you're right that it is the standard!

But alas, according to your link, "[there is no perfect email regex](http://www.regular-expressions.info/email.html)", which matches up with what I remember having read about email regexes in the past in that not _all_ issued email addresses follow the standard, so trying to validate email addresses against it is probably not the best idea. :)

tools/reviewer/web/src/app/dashboard/diff/diff-discussion/diff-discussion.component.ts
webdevelopland: It's 331.6 kB, but ok.

tools/reviewer/web/src/app/dashboard/diff/diff-header/diff-header-content/diff-header-content.component.ts
webdevelopland: It's 331.6 kB, but ok

PR #221, opened 2018-08-03
Integrate formatting check into `aa fix`

vmax:


tools/aa/commands/FixCommand.java
```diff
@@ -20,15 +20,26 @@
...
 public class FixCommand implements AaCommand {
 
-  ImmutableList<FixCommandCheck> checks =
-      ImmutableList.of(new DummyCheck(), new FailingDummyCheck());
+  List<FixCommandCheck> checks;
 
   @Inject
-  public FixCommand() {}
+  public FixCommand(
+      Lazy<FormattingCheck> formattingCheck,
```
oferb: `Lazy` is helpful if you don't always want to instantiate the object, and so you can choose to do so in `.get()`.
Since here you're instantiating always, `Lazy` is not useful, and you can just inject without it.

tools/aa/commands/FixCommand.java
```diff
@@ -20,15 +20,26 @@
...
+      Lazy<DummyCheck> dummyCheck,
+      Lazy<FailingDummyCheck> failingDummyCheck) {
+    ImmutableList.Builder builder = ImmutableList.builder();
+    builder.add(formattingCheck.get());
```
oferb: Do we still need `DummyCheck` and `FailingDummyCheck`?

tools/aa/commands/checks/FormattingCheck.java
```diff
@@ -0,0 +1,79 @@
...
+              Files.walk(Paths.get(""))
+                  .filter(Files::isDirectory)
+                  .filter(p -> p.getFileName().toString().equals("node_modules"))
+                  .map(p -> p.toAbsolutePath().toString())
+                  .collect(Collectors.toList()));
+    } catch (IOException ex) {
+      throw new RuntimeException(ex);
+    }
+
+    System.err.println("curdir: " + this.workspacePath);
```
oferb: What does this println do?

tools/aa/commands/checks/FormattingCheck.java
```diff
@@ -0,0 +1,79 @@
...
+
+    FormatterTool.main(
```
oferb: Can you add the following TODO?
`// TODO: Move formatting logic from FormatterTool to a new Formatter class.`

tools/aa/commands/FixCommand.java
vmax: @oferb nope, they only served as an example

tools/aa/commands/checks/FormattingCheck.java
vmax: @oferb oops, that was a debug one, I should have been more attentive

PR #231, opened 2018-08-09
Organize tools and documentation

oferb:
Specifically, tools/deps, examples/android and split compile.sh into build and test.


tools/deps/update_maven_deps.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
val-fed: I dot exception when to run it:
```
[main] ERROR MakeDeps - Failed to read dependencies.yaml
java.io.FileNotFoundException: /home/valerii/Work/review/D150/startup-os/tools/deps/dependencies.yaml (No such file or directory)
```
Also in `tools/deps/add_maven_deps.sh`

val-fed: This comment isn't actual. Sorry!

PR #241, opened 2018-08-11
Whitelist for generated package-lock.bzl

vmax:


tools/deps/PackageLockWhitelistTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+  @Before
+  public void setUp() throws Exception {
+    whitelist = new Yaml().load(new FileInputStream(new File("tools/deps/whitelist.yaml")));
+    packageLockLines = Files.readAllLines(Paths.get("third_party/maven/package-lock.bzl"));
+    parsedDependencies = new ArrayList();
+  }
+
+  @Test
+  public void parseDependencies() throws Exception {
```
oferb: You seem to assume `parseDependencies()` will be called before `validateDependencies()`, which may not be the case in parallel runners.

tools/deps/PackageLockWhitelistTest.java
```diff
@@ -0,0 +1,95 @@
...
+      if (line.equals("def list_dependencies():")) {
+        depListStarted = true;
+        continue;
+      }
+
+      if (line.equals("def maven_dependencies(callback = declare_maven)")) {
+        break;
+      }
+
+      if (depListStarted && line.contains("://")) {
```
oferb: Why do you need `depListStarted`? You can check `://` on all lines.
Why break on `def maven_dependencies(callback = declare_maven)` specifically, if it doesn't contain `://`?

tools/deps/PackageLockWhitelistTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+  @Test
+  public void validateDependencies() throws Exception {
+    assertFalse("Parsed dependencies list should not be empty", parsedDependencies.isEmpty());
+
+    List<String> validPackageGroups = whitelist.get("maven_dependencies");
+    for (Dependency dep : parsedDependencies) {
+
+      assertEquals(
+          "Artifact %s is not in the Maven Central", MAVEN_CENTRAL_URL, dep.getRepository());
```
oferb: s/in the/in

tools/deps/PackageLockWhitelistTest.java
```diff
@@ -0,0 +1,87 @@
...
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+import org.junit.Before;
+import org.junit.Test;
+import org.yaml.snakeyaml.Yaml;
+
+public class PackageLockWhitelistTest {
```
oferb: Can you document what the test checks and what it assumes?

PR #252, opened 2018-08-15
GitRepo: Set original_filename for RENAME and COPY

val-fed:


common/repo/tests/GitRepoTest.java
```diff
@@ -249,6 +249,24 @@ public void testGetUncommittedFilesWhenDeletedFile() {
...
   }
 
+  @Test
+  public void testGetUncommittedFilesWhenRenamedFile() {
+    repo.switchBranch(TEST_BRANCH);
+    fileUtils.writeStringUnchecked(
+        TEST_FILE_CONTENTS, fileUtils.joinToAbsolutePath(repoFolder, TEST_FILE));
+    gitRepo.addFile(TEST_FILE);
+    repo.commit(repo.getUncommittedFiles(), COMMIT_MESSAGE);
+    gitRepo.renameOrMove(TEST_FILE, "new_name.txt");
```
oferb: Why move the file through `git mv` and not directly through the FileSystem?

val-fed: If the file is moved directly, Git shows:
- the old file as `DELETED`
- new file as `ADDED`
If `git mv` command is used, Git shows:
- only one file as `RENAMED`

oferb: If you add the files to staging (`git add`) it shows them as renamed. The files are unstaged, right? Maybe we should stage them instead of using `git mv`, since most people usually just move files using the filesystem (I think).

PR #253, opened 2018-08-16
Disallow copying node modules into workspace

val-fed:


common/FileUtils.java
```diff
@@ -183,7 +186,7 @@ public String getCurrentWorkingDirectoryName() {
             fileSystem.getPath(expandHomeDirectory(path)),
             100000, // Folder depth
             (unused, unused2) -> true)) {
-      return ImmutableList.sortedCopyOf(paths.map(Path::toString).collect(Collectors.toList()));
+      return ImmutableList.sortedCopyOf(paths.map(Path::toString).collect(toList()));
```
oferb: Let's keep `Collectors` - that way it's clear it's not a local method.

common/FileUtils.java
```diff
@@ -238,19 +241,36 @@ public Message readProtoBinaryUnchecked(String path, Message.Builder builder) {
    * Copies a directory to another directory. Creates target directory if needed and doesn't copy
    * symlinks.
    */
-  public void copyDirectoryToDirectory(String source, String destination, String ignored)
+  public void copyDirectoryToDirectory(String source, String destination, String... ignored)
       throws IOException {
     final Path sourcePath = fileSystem.getPath(source);
     final Path targetPath = fileSystem.getPath(destination);
+
+    List<String> allFilesAndFoldersForIgnore = new ArrayList<>();
```
oferb: s/allFilesAndFoldersForIgnore/filesAndFoldersToIgnore

common/FileUtils.java
```diff
@@ -238,19 +241,36 @@ public Message readProtoBinaryUnchecked(String path, Message.Builder builder) {
...
+    if (ignored.length != 0) {
+      for (String itemForIgnore : ignored) {
+        String itemForIgnorePath = joinPaths(source, itemForIgnore);
+        if (folderExists(itemForIgnorePath) || fileExists(itemForIgnorePath)) {
+          allFilesAndFoldersForIgnore.addAll(getListContentsWithAbsolutePath(itemForIgnorePath));
```
oferb: Why do we need `getListContentsWithAbsolutePath()`? Can you give an example?

tools/aa/commands/WorkspaceCommand.java
```diff
@@ -89,7 +89,10 @@ private boolean createWorkspace() {
       fileUtils.mkdirs(workspacePath);
       try {
         fileUtils.copyDirectoryToDirectory(
-            fileUtils.joinPaths(config.getBasePath(), "head"), workspacePath, "^bazel-.*$");
+            fileUtils.joinPaths(config.getBasePath(), "head"),
+            workspacePath,
+            "^bazel-.*$",
+            "startup-os/tools/local_server/web_login/node_modules");
```
oferb: Let's ignore all `node_modules`, so we don't need to remember to update this list.

common/FileUtils.java
jbduncan: In my personal opinion, `collect(toList())` reads better because of how it looks like prose. (I'm also reasonably sure that's how the OpenJDK devs intended `Collectors` to be used.) But I won't insist. :)

jbduncan: Hmm, actually, I realise that the result of `collect(toList())` is being passed to a call to `ImmutableList.sortedCopyOf`.

Thus I wonder if the following would read better and be equally performant:

```java
return paths.map(Path::toString).sorted().collect(toImmutableList());
```

What do you think?

common/FileUtils.java
val-fed: I deleted this method and simplify the code.

common/FileUtils.java
val-fed: Deleted it.

tools/aa/commands/WorkspaceCommand.java
val-fed: Done.

tools/aa/commands/WorkspaceCommand.java
```diff
@@ -101,6 +107,22 @@ private boolean createWorkspace() {
     return true;
   }
 
+  /** Returns list from head folder with all path where file or folder to ignore is found */
+  private ImmutableList<String> getFilesOrFoldersToIgnoreInHead(String fileOrFolderToIgnore)
```
oferb: This would not be efficient if `node_modules` has many files. We would like to ignore `node_modules` without listing att files inside.

common/FileUtils.java
```diff
@@ -243,19 +243,28 @@ public Message readProtoBinaryUnchecked(String path, Message.Builder builder) {
...
     Files.walkFileTree(
         sourcePath,
         new SimpleFileVisitor<Path>() {
           @Override
           public FileVisitResult preVisitDirectory(final Path dir, final BasicFileAttributes attrs)
               throws IOException {
-            if (ignored != null) {
-              if (Pattern.matches(ignored, dir.getFileName().toString())) {
-                return FileVisitResult.CONTINUE;
+            final String dirPath = dir.toAbsolutePath().toString();
```
oferb: Aha! We are iterating over all files, and we were before as well.
If a folder is ignored, we should return `FileVisitResult.SKIP_SUBTREE`.
I hope this speeds things up.

common/FileUtils.java
```diff
@@ -265,9 +274,17 @@ public FileVisitResult preVisitDirectory(final Path dir, final BasicFileAttribut
           @Override
           public FileVisitResult visitFile(final Path file, final BasicFileAttributes attrs)
               throws IOException {
-            if (ignored != null) {
-              if (Pattern.matches(ignored, file.getFileName().toString())) {
-                return FileVisitResult.CONTINUE;
+            final String filePath = file.toString();
```
oferb: Let's only ignore at the folder level. If it's a file - we always copy it.

tools/aa/commands/WorkspaceCommand.java
```diff
@@ -88,10 +91,13 @@ private boolean createWorkspace() {
     } else {
       fileUtils.mkdirs(workspacePath);
       try {
+        List<String> filesAndFoldersToIgnore = new ArrayList<>();
+        filesAndFoldersToIgnore.add("^bazel-.*$");
+        filesAndFoldersToIgnore.addAll(getFilesOrFoldersToIgnoreInHead("node_modules"));
```
oferb: What if we do a regex?
Try:
`node_modules`
`^node_modules$`
Not sure if we're matching against the full path or the folder name.

tools/aa/commands/WorkspaceCommand.java
val-fed: This method doesn't return all folders and files inside the `file_or_folder_to_ignore`. It just founds all paths with where is `file_or_folder_to_ignore`. For example, we have the following:
```
/head/repo1/a_lot_of_folders_and_files
/head/repo1/path/to/node_modules
/head/repo1/another/path/to/node_modules

/head/repo2/a_lot_of_folders_and_files
/head/repo2/some/path/to/node_modules
```
This method will return folloving list:
- `/head/repo1/path/to/node_modules`
- `/head/repo1/another/path/to/node_modules`
- `/head/repo2/some/path/to/node_modules`

common/FileUtils.java
val-fed: Thanks! It's a really good option.

common/FileUtils.java
val-fed: Done.

tools/aa/commands/WorkspaceCommand.java
val-fed: Added regex support.

tools/aa/commands/WorkspaceCommand.java
val-fed: I deleted this method. The last comment isn't actual.

common/FileUtils.java
oferb: Valerii, can you check the shorter version? It looks nice.
About removing the `Collectors.`, I think that the convention of methods not bring imported has the benefit that when you see a method, you don't need to think twice to know that it's local.
I can see the benefits of shortening the code, but once you do it here (and maybe in a few other places here and there), you lose that confidence across the whole codebase.

val-fed: I tried to use the shorter version:
 ```
return paths.map(Path::toString).sorted().collect(toImmutableList());
```
`toImmutableList()` required additional import:
```
import static autovalue.shaded.com.google$.common.collect.$ImmutableList.toImmutableList;
```
After these changes, `listContentsRecursively()` method will return `autovalue.shaded.com.google$.common.collect.$ImmutableList` instead `com.google.common.collect.ImmutableList`. 

And it requires changing all the code where `listContentsRecursively()` method is used.

PR #255, opened 2018-08-20
Make aa init work without base_path flag

val-fed:


tools/aa/commands/InitCommand.java
```diff
@@ -96,11 +98,34 @@ public boolean run(String[] args) {
...
+  private boolean processArgs(String[] args) {
+    if (args.length == 0 || args.length == 1) {
+      System.err.println(
+          RED_ERROR
+              + "Invalid usage. \n"
+              + "Available variants are: \"bazel run //tools/aa:aa_tool -- init </path/to/base/folder>\" "
+              + "or, if aa is already set up: \"aa init <base_path>\"");
+      return false;
+    }
+    boolean aaIsAlreadySetUpMode = args[0].equals("aa");
```
oferb: s/aaIsAlreadySetUpMode/isAaSetUp

tools/aa/commands/InitCommand.java
```diff
@@ -96,11 +98,34 @@ public boolean run(String[] args) {
     return true;
   }
 
+  private boolean processArgs(String[] args) {
+    if (args.length == 0 || args.length == 1) {
```
oferb: Let's check if `aa` is set up before this message, and then print the relevant part of the message.

tools/aa/commands/InitCommand.java
val-fed: Done.

tools/aa/commands/InitCommand.java
val-fed: Done it. I also added a `switch` instead of `if else` constructions. It should be more understandable.

tools/reviewer/service/tests/CodeReviewServiceTextDiffTest.java
```diff
@@ -174,11 +174,10 @@ private void createInitialRepo(String initialRepoFolder) {
 
   private void initAaBase(String initialRepoFolder, String aaBaseFolder) {
     InitCommand initCommand = component.getInitCommand();
-    InitCommand.basePath.resetValueForTesting();
     InitCommand.startuposRepo.resetValueForTesting();
     String[] args = {
+      "init", aaBaseFolder,
```
oferb: If we removed `--base_path`, why do we need to add `init`?

tools/aa/commands/InitCommand.java
```diff
@@ -95,11 +97,41 @@ public boolean run(String[] args) {
...
+    if (isAaSetUp) {
+      switch (args.length) {
+        case 1:
+          System.err.println(
+              RED_ERROR
+                  + "Invalid usage. \n"
+                  + "Please use \"aa init <base_path>\" command to init a base folder.");
+          return false;
+        case 2:
+          System.err.println(RED_ERROR + "Missing base_path");
```
oferb: I think you should end in `ANSI_RESET` to prevent the next text to be red too.

tools/aa/commands/InitCommand.java
```diff
@@ -95,11 +97,41 @@ public boolean run(String[] args) {
     return true;
   }
 
+  private boolean processArgs(String[] args) {
+    boolean isAaSetUp = args[0].equals("aa");
```
oferb: Sorry, I think this actually is better:
s/isAaSetUp/isAaAlreadySetUp

tools/aa/commands/InitCommand.java
```diff
@@ -95,11 +97,41 @@ public boolean run(String[] args) {
...
+          return false;
+        case 2:
+          System.err.println(RED_ERROR + "Missing base_path");
+          return false;
+        default:
+          basePath = args[2];
+          break;
+      }
+    } else {
+      switch (args.length) {
```
oferb: Instead of duplicating, how about you remove the first element if needed, using:
`Arrays.copyOfRange(Object[] src, int from, int to)`
Then, you can use the same code?

tools/reviewer/service/tests/CodeReviewServiceTextDiffTest.java
val-fed: It happens because we changed the logic of parsing input args.
We used flags before:
- `bazel run //tools/aa:aa_tool -- init --base_path </path/to/base/folder> --startupos_repo <url_to_startupos_repo>`
- `aa init --base_path </path/to/base/folder> --startupos_repo <url_to_startupos_repo>`
And if we use `init` command directly (`initCommand.run(args)`) we could just to supply flags in an arbitrary manner.

Now, init command looks like:
- `bazel run //tools/aa:aa_tool -- init </path/to/base/folder> --startupos_repo <url_to_startupos_repo>`
- `aa init </path/to/base/folder> --startupos_repo <url_to_startupos_repo>`
Now, input args is parsed otherwise. We get </path/to/base/folder> by a specific index (args[1] or args[2]). And if <path/to/base/folder> will be not in the right place it won't be working.

tools/aa/commands/InitCommand.java
val-fed: Do you mean `System.err.println(RED_ERROR + "Missing base_path" + ANSI_RESET)`?

I don't see any diff beetwen `(RED_ERROR + "Missing base_path" + ANSI_RESET)` and `(RED_ERROR + "Missing base_path").` In both cases, the text looks like:
<red>ERROR:<red> <usual>Missing base_path<usual>
<usual>some text<usual>

tools/aa/commands/InitCommand.java
val-fed: - [x] Done

tools/aa/commands/InitCommand.java
val-fed: - [x] Done

tools/aa/commands/InitCommand.java
```diff
@@ -95,11 +98,35 @@ public boolean run(String[] args) {
...
+                + "Invalid usage. \n"
+                + "Please use \"aa init <base_path>\" command to init a base folder.");
+        return false;
+      }
+      // leave in `initArgs` array only `init` and `</path/to/base/folder>` arguments
+      initArgs = Arrays.copyOfRange(initArgs, 1, initArgs.length);
+    }
+    if (initArgs.length == 1) {
+      System.err.println(RED_ERROR + "Missing base_path" + ANSI_RESET);
+      System.err.println("some text1");
```
oferb: Probably forgot these println's.

tools/aa/commands/InitCommand.java
```diff
@@ -95,11 +98,35 @@ public boolean run(String[] args) {
     return true;
   }
 
+  private boolean processArgs(String[] args) {
+    String[] initArgs = Arrays.copyOf(args, args.length);
+    boolean isAaAlreadySetUp = initArgs[0].equals("aa");
```
oferb: Let's add:
// aa init can be called by either `bazel run //tools/aa:aa_tool -- init` or `aa init`. For `aa init`, there is an extra first argument `aa` in `args`.

Once you do that, I think you don't need `isAaAlreadySetUp` anymore, you can just put it directly inside the `if`.

tools/aa/commands/InitCommand.java
```diff
@@ -95,11 +98,35 @@ public boolean run(String[] args) {
...
+    if (isAaAlreadySetUp) {
+      if (initArgs.length == 1) {
+        System.err.println(
+            RED_ERROR
+                + "Invalid usage. \n"
+                + "Please use \"aa init <base_path>\" command to init a base folder.");
+        return false;
+      }
+      // leave in `initArgs` array only `init` and `</path/to/base/folder>` arguments
+      initArgs = Arrays.copyOfRange(initArgs, 1, initArgs.length);
```
oferb: Let's add:
// We remove the first element so that both calls to `aa init` have the same args.

I think once we add that, we can just reuse `args`. It's a reference anyway, so you're not changing the real array.

tools/aa/commands/InitCommand.java
val-fed: Yes, sorry. I forgot about it.

tools/aa/commands/InitCommand.java
val-fed: Done.

tools/aa/commands/InitCommand.java
val-fed: Done.

tools/aa/commands/InitCommand.java
```diff
@@ -95,11 +98,33 @@ public boolean run(String[] args) {
...
+            RED_ERROR
+                + "Invalid usage. \n"
+                + "Please use \"aa init <base_path>\" command to init a base folder.");
+        return false;
+      }
+      // We remove the first element so that both calls to `aa init` have the same args
+      args = Arrays.copyOfRange(args, 1, args.length);
+    }
+    if (args.length == 1) {
+      System.err.println(RED_ERROR + "Missing base_path" + ANSI_RESET);
```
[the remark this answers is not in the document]
val-fed: Done.

PR #278, opened 2018-09-09
Reviewer - resolve comments

webdevelopland:
What's new:
- Multi-threading. Ability to leave several (more than one) threads under 1 line.
- Functionality to `resolve` a thread. Resolved threads aren't displayed on a screen.
- Small cosmetic changes

I recommend you to review the commits separately to not be confused.


MateMahinKhan:
If in future you end up writing more code for handling time and formats, I recommend you use moment.js. Less code in the app would mean less maintenance and less possible points of failure. Dates are a complex thing that needs a lot of ifs and elses.

PR #282, opened 2018-09-10
Created and updated timestamps for Diffs

vmax:


tools/reviewer/aa/commands/DiffCommand.java
```diff
@@ -97,7 +97,8 @@ private Diff createDiff() {
             .setDescription(description.get())
             .setBug(buglink.get())
             .addAllReviewer(getReviewers(reviewers.get()))
-            .setId(response.getLastDiffId());
+            .setId(response.getLastDiffId())
+            .setCreatedTimestamp(new Long(System.currentTimeMillis()));
```
oferb: I guess we should set both `setCreatedTimestamp` and `setModifiedTimestamp` here.

Then setting it in `ReviewCommand.java` becomes redundant.

PR #284, opened 2018-09-10
Improve aa options autocomplete for add_repo, diff

vmax:


oferb:
[Check aa add_repo to make sure it works, add an example in the error message and in the documentation, and add it to the auto-complete.](https://trello.com/c/8WdNjGMc/4-check-aa-addrepo-to-make-sure-it-works-add-an-example-in-the-error-message-and-in-the-documentation-and-add-it-to-the-auto-compl)

PR #319, opened 2018-09-15
Create global registry for Reviewer instances

oferb:
This PR also adds a Reviewer instance config.

### DIFF
--- a/.reviewer/README.md
+++ b/.reviewer/README.md
@@ -1,0 +1,14 @@
+# Reviewer config
+This config contains a global registry of Reviewer instances, as well as configuration for this
+specific instance.
+
+# Global Registry
+A list of Reviewer instances. List is edited as prototxt and then stored as protobin to avoid
+breaking older versions. A test ensures the prototxt and protobin match and prints out the updated
+protobin upon change.
+
+# Config for this instance
+The config includes:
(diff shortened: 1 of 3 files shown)


.reviewer/README.md
```diff
@@ -0,0 +1,14 @@
+# Reviewer config
+This config contains a global registry of Reviewer instances, as well as configuration for this
+specific instance.
+
+# Global Registry
+A list of Reviewer instances. List is edited as prototxt and then stored as protobin to avoid
+breaking older versions. A test ensures the prototxt and protobin match and prints out the updated
```
vmax: @oferb 
>A test ensures the prototxt and protobin match

How will you ensure that test is ran on every proto update? Generally, will it be updated via UI or modifying proto text, running that test and overwriting protobin manually?

oferb: It will updated by modifying proto text, running that test and overwriting protobin manually.
You can still do it in the GitHub UI though. You would modify, see CircleCI fail, copy the protobin, and then it would succeed and you can merge. So you don't need to run it on your local machine.

vmax: @oferb Given this file is unlikely to change too often, I think it's OK. Approved.

PR #321, opened 2018-09-15
aa killserver command impl; fixes #303

### LINKED ISSUE #303: Add command: aa killserver
The command essentially does this:
```
# Kill previous:
if [ "$(uname)" = "Darwin" ] || [ "$(uname)" = "Linux" ]; then
  # Mac & Linux
  # kill HTTP gateway (port 7000), gRPC server (port 8001) and Angular server (8000)
  # -t makes `lsof` output only PIDs so output can be piped to `kill`
  # -n and -P prevent `lsof` from resolving addresses and ports, therefore
  # making execution faster
  # we filter for processes that listen on port in order to kill only
  # servers (i.e. Angular) instead of clients (i.e. Chrome)
  kill $(lsof -tnP -i:7000 -i:8000 -i:8001 -sTCP:LISTEN)
# else
#   # Windows
#   # TODO: kill local_http_gateway & local_server
fi
```

vmax:


tools/reviewer/local_server/run.sh
```diff
@@ -1,43 +1,60 @@
 # Run LocalHttpGateway and LocalServer
 # Usage:
-# run.sh --root_path <root_path>
+# run.sh start --root_path <root_path>
```
oferb: Let's rename `run.sh` to `local_server.sh`, since `run.sh stop` seems confusing.

PR #338, opened 2018-09-22
Reviewer-job to sync global Reviewer instances registry

vmax:


tools/reviewer/job/impl/ReviewerMetadataUpdaterTask.java
```diff
@@ -0,0 +1,182 @@
...
+                      globalRegistryFilePath, ReviewerRegistry.newBuilder());
+
+          uploadProtoToFirestore(reg);
+        } else {
+
+          String newChecksum = md5ForFile(globalRegistryFilePath);
+          if (!newChecksum.equals(storedChecksum)) {
+            log.atInfo()
+                .log(
+                    "Checksum not equal to stored: found %s, expected %s",
```
elinabtl: How about, `New checksum not equal to stored one: new %s, stored %s`

tools/reviewer/job/impl/ReviewerMetadataUpdaterTask.java
```diff
@@ -0,0 +1,182 @@
...
+                        globalRegistryFilePath, ReviewerRegistry.newBuilder());
+
+            uploadProtoToFirestore(reg);
+
+            storedChecksum = newChecksum;
+          } else {
+            log.atInfo().log("Checksum equals to stored: found %s,", storedChecksum);
+          }
+        }
+      } catch (IOException ignored) {
```
elinabtl: Let's log it though..

tools/reviewer/job/Job.java
```diff
@@ -0,0 +1,61 @@
...
+  @Singleton
+  @Component(modules = {CommonModule.class})
+  public interface JobComponent {
+    Job getJob();
+  }
+
+  private void run(String[] args) {
+    Flags.parseCurrentPackage(args);
+
+    new Timer().scheduleAtFixedRate(this.taskExecutor, 0, 1000L);
```
elinabtl: Would be nice to be able to have every task run at different periodicity, and choose whether it:
* Wants overlapping runs
* Wants to skip overlapping runs
* Wants to run immediately after overlapping run ends
* Wants to run some time after previous run ended
We don't have to implement that now, but those are the main scenarios I think we'll encounter.
It would be nice it the task would choose the strategy and parameters, and some common code would take care of it.

You can add a TODO for it.

tools/reviewer/job/Job.java
```diff
@@ -0,0 +1,61 @@
...
+  }
```
elinabtl: I think querying every second is too frequent. Let's do it every 15 seconds.

tools/reviewer/job/impl/ReviewerMetadataUpdaterTask.java
```diff
@@ -0,0 +1,182 @@
...
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+public class ReviewerMetadataUpdaterTask implements Task {
```
elinabtl: Can you describe what the task does?

tools/reviewer/job/impl/ReviewerMetadataUpdaterTask.java
```diff
@@ -0,0 +1,182 @@
...
+
+          ReviewerRegistry reg =
+              (ReviewerRegistry)
+                  fileUtils.readPrototxtUnchecked(
+                      globalRegistryFilePath, ReviewerRegistry.newBuilder());
+
+          uploadProtoToFirestore(reg);
+        } else {
+
+          String newChecksum = md5ForFile(globalRegistryFilePath);
```
elinabtl: If we have a checksum, we should probably still `repo.pull();` right?

tools/reviewer/job/impl/ReviewerMetadataUpdaterTask.java
```diff
@@ -0,0 +1,182 @@
...
+          if (fileUtils.folderEmptyOrNotExists(REPO_DIRECTORY)) {
+            repo.cloneRepo(Job.repoUrl.get(), REPO_DIRECTORY);
+          } else {
+            repo.pull();
+          }
+
+          storedChecksum = md5ForFile(globalRegistryFilePath);
+          log.atInfo().log("Computed checksum: %s", storedChecksum);
+
+          ReviewerRegistry reg =
```
elinabtl: I think we can simplify this function by merging the main if-else into a single flow:
* Make sure file is fresh, pull/clone as needed
* Get checksum
* If not equal to stored, whether because stored is null or not
* Do the main action and update stored

PR #349, opened 2018-10-05
Hasadna CI in startup-os

vmax:


tools/deps/WorkspacePatcher.java
```diff
@@ -0,0 +1,50 @@
...
+
+import com.google.startupos.common.flags.Flags;
+import com.google.startupos.common.flags.Flag;
+import com.google.startupos.common.flags.FlagDesc;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.io.IOException;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+// TODO: refactor if more advanced functions/extension is needed
```
oferb: I think we don't need this comment :)
The code is fine, does what it needs, is simple. We don't need to apologize :)

Please describe what the class does though :)

.circleci/config.yml
```diff
@@ -70,6 +70,29 @@ jobs:
...
+
+      - run: sudo cp .circleci/bazel.rc /etc/bazel.bazelrc
+      - run: bazel info release
+      - run: './tools/get_android_sdk.sh'
+      - run:
+          command: git clone https://github.com/hasadna/hasadna /home/circleci/hasadna/
+      - run:
+          command: bazel run //tools/deps:workspace_patcher -- --workspace /home/circleci/hasadna/WORKSPACE
+      - run:
+          command: cd /home/circleci/hasadna/ && ./compile.sh build
```
oferb: Why not `./compile.sh test`?

vmax: @oferb I think we need both 🙂

Issue #360, opened 2018-10-13
.circleci/config.yml error


lloydchang:
```
Config Processing Error

$ #!/bin/sh -eo pipefail
  # Unable to parse YAML
  # while constructing a mapping
  #  in 'string', line 64, column 11:
  #               key: \"android_sdk\"
  #               ^
  # found duplicate key key
  #  in 'string', line 67, column 11:
  #               key: \"v3-bazel_cache\"
  #               ^
  false

Exited with code 1
```
```
$ circleci config validate
Error: Unable to parse YAML: while constructing a mapping:  in 'string', line 64, column 11::               key: "android_sdk":
    ^: found duplicate key key:  in 'string', line 67, column 11::               key: "v3-bazel_cache":
```

vmax:
Fixed in 98780f0

PR #361, opened 2018-10-13
resolve YAML error #360

lloydchang:


vmax:
@lloydchang thank you for your contribution! CI didn't fail with duplicate keys though, but as I looked through the docs the didn't specify it's OK either, so I guess it's safe to merge

PR #372, opened 2018-10-16
GitHub Sync: Fix GithubReader

val-fed:
Set `repoId`, `file`,  `type` only once, not to override these values for each comment.

### DIFF
--- a/tools/reviewer/job/sync/GithubReader.java
+++ b/tools/reviewer/job/sync/GithubReader.java
@@ -195,32 +195,31 @@ public class GithubReader {
                 .filter(comment -> comment.getPosition() == position)
                 .collect(Collectors.toList()));
 
-    Thread.Builder thread = Thread.newBuilder();
+    Thread.Builder thread =
+        Thread.newBuilder()
+            .setRepoId(pr.getRepoName())
+            .setFile(
+                Protos.File.newBuilder()
+                    .setFilename(positionComments.get(0).getPath())
+                    .setRepoId(pr.getRepoName())
+                    .setFilenameWithRepo(
+                        pr.getHead().getRepo().getName() + "/" + positionComments.get(0).getPath())
+                    .setCommitId(positionComments.get(0).getCommitId())
+                    // TODO: Think over how to save already received user's email from the
+                    // previous iteration. It can reduce the number of requests.
+                    .setUser(
+                        githubClient
+                            .getUser(
+                                UserRequest.newBuilder().setLogin(pr.getUser().getLogin()).build())
+                            .getUser()
+                            .getEmail())
+                    .build())
+            .setType(Thread.Type.CODE);
+
     positionComments.forEach(
         comment ->
             thread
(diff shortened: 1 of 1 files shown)


tools/reviewer/job/sync/GithubReader.java
```diff
@@ -195,32 +195,31 @@ private Thread getCodeThreadByFilenameAndPosition(
...
+                    .build())
+            .setType(Thread.Type.CODE);
+
     positionComments.forEach(
         comment ->
             thread
-                // TODO: Set `repoId`, `file`, `lineNumber`, `type` only once, not to override these
-                // values for each comment.
-                .setRepoId(pr.getRepoName())
                 .setCommitId(setThreadCommitId(pr, comment))
```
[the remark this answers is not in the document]
val-fed: You are right. Fixed it.

tools/reviewer/job/sync/GithubReader.java
```diff
@@ -195,53 +195,57 @@ private Thread getCodeThreadByFilenameAndPosition(
                 .filter(comment -> comment.getPosition() == position)
                 .collect(Collectors.toList()));
 
-    Thread.Builder thread = Thread.newBuilder();
+    GithubComment firstComment;
+    if (positionComments.isEmpty()) {
+      throw new RuntimeException("Can not find comments for a position: " + position);
```
oferb: s/Can not/Can't

tools/reviewer/job/sync/GithubReader.java
```diff
@@ -195,53 +195,57 @@ private Thread getCodeThreadByFilenameAndPosition(
...
+    } else {
+      firstComment = positionComments.get(0);
```
oferb: You don't need the else - you can just;
`GithubComment firstComment = positionComments.get(0);`

tools/reviewer/job/sync/GithubReader.java
```diff
@@ -195,53 +195,57 @@ private Thread getCodeThreadByFilenameAndPosition(
                 .filter(comment -> comment.getPosition() == position)
                 .collect(Collectors.toList()));
 
-    Thread.Builder thread = Thread.newBuilder();
+    GithubComment firstComment;
```
oferb: Please add a comment for why you use the first comment - that you are trying to fit several comments into a thread, and you assume all comments have the same values for these fields (is that really true?)

tools/reviewer/job/sync/GithubReader.java
val-fed: Done.

tools/reviewer/job/sync/GithubReader.java
val-fed: Done.

tools/reviewer/job/sync/GithubReader.java
val-fed: Done.

Issue #374, opened 2018-10-16
GitHub Sync: Add LineNumberConverter tests


val-fed:
Add LineNumberConverter tests to make sure converting `patch position` <--> `file line number` is going rightly.

PR #375, opened 2018-10-16
GitHub Sync: Add LineNumberConverter tests

val-fed:
#374

### DIFF
--- a/tools/reviewer/job/sync/tests/BUILD
+++ b/tools/reviewer/job/sync/tests/BUILD
@@ -5,7 +5,6 @@ java_test(
     srcs = [
         "LineNumberConverterTest.java",
     ],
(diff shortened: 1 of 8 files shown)


tools/reviewer/job/sync/tests/BUILD
```diff
@@ -0,0 +1,15 @@
+package(default_visibility = ["//visibility:public"])
+
+java_test(
+    name = "line_number_converter_test",
+    srcs = [
+        "LineNumberConverterTest.java",
+    ],
+    local = 1,
```
oferb: Why `local = 1`?

val-fed: Deleted it.

PR #383, opened 2018-10-20
webapp - ui changes 3

webdevelopland:
What's new:
- new `login page` design
- app modules refactoring
- approved users are highlighted green
- functionality to Expand/Collapse threads
- `No Discussion yet`, `No Files yet`


tools/reviewer/webapp/src/app/dashboard/diff/diff-discussion/discussion.service.ts
```diff
@@ -36,13 +36,13 @@ export class DiscussionService {
       .length;
   }
 
-  // If new thread wasn't added or a thread wasn't deleted make links update.
+  // If new thread wasn't added or a thread wasn't deleted then make "links update".
```
oferb: What is "links update"? Please explain in the comment.

tools/reviewer/webapp/src/app/dashboard/diff/diff-header/diff-header-content/diff-header-content.component.html
```diff
@@ -1,6 +1,6 @@
 <div class="header-content cr-header">
   <!--
-    Block with main information of the diff
+    This's a block with main information of the diff
```
oferb: s/This's/This is
We don't use `This's` as a shortening of `This is`.

tools/reviewer/webapp/src/app/shared/services/auth.guard.ts
```diff
@@ -1,31 +1,15 @@
...
 export class AuthGuard implements CanActivate {
-  constructor(private router: Router, private authService: AuthService) { }
+  constructor(private angularFireAuth: AngularFireAuth) { }
 
-  canActivate(
-    route: ActivatedRouteSnapshot,
-    state: RouterStateSnapshot,
-  ): Observable<boolean> {
-    // There is a bug.
-    // If you open the app without internet connection,
```
oferb: I tried this, and got the Chrome dinosaur, so now sure how to reproduce. Is this resolved now?

tools/reviewer/webapp/src/app/dashboard/diff/diff-discussion/discussion.service.ts
webdevelopland: ?????????????????????

MateMahinKhan:
Having feature modules is very important. You just have import the shared module and into your feature modules. Feature modules compartmentalize the app. They give sort-of a boundary between features. One huge Root Module will look ugly. That's why angular itself has many different modules like FormsModule and Angular material project, that can be taken as an example, has a separate module for each component.
Imagine more than one developer working on the module file and you will have merge conflicts every time even if you guys are working on separate features.
Having different modules helps us in lazyloading them and that's a huge feature of Angular. You don't have to load the whole app anytime somebody uses only a part of application.
Here are some relevant links:
https://angular.io/guide/feature-modules
https://mirkomaggioni.com/2017/02/12/feature-modules-in-angular-2/ 
That being said; If you believe the reviewer web app is still not big enough to have feature modules and benefit from lazyloading you can have everything in the root module but remember there's nothing as permanent as a temporary solution ;) 
I'm approving the PR because everything looks good if we're going ahead with a root module.

tools/reviewer/webapp/src/app/dashboard/diff/diff-discussion/discussion.service.ts
webdevelopland: :upside_down_face:

webdevelopland: But there's already the explanation on the next line

webdevelopland: :upside_down_face:

webdevelopland: But there's already the explanation on the next line
https://i.imgur.com/FVVq76Q.png

tools/reviewer/webapp/src/app/shared/services/auth.guard.ts
webdevelopland: To reproduce it you need to start app from source.
But bug is gone. Not by the update. Somewhere before, I'm not sure what is moment it's.

PR #392, opened 2018-10-30
GitHub Sync: Add DiffConverter, DiffReader and sync logic

val-fed:


tools/reviewer/job/sync/GithubSync.java
```diff
@@ -56,53 +44,263 @@
...
-          {
-            githubSync.readPullRequest(githubClient);
-            break;
+    // Go over all Reviewer Diff numbers to sync
+    for (long diffNumber : diffsToSync.get()) {
+      syncWithGithub(githubClient, diffNumber);
+    }
+  }
+
+  private static void syncWithGithub(GithubClient githubClient, long diffNumber)
```
oferb: Generally when a method becomes very long - we have 2 main strategies to cope with it:
* Extract repetitive parts to another method, sometimes with a parameter that changes the code to fit the different cases.
* Extract non-repetitive parts to another method, usually breaking the code into parts. For example, one part could be left in the original method, and the other part in the extracted method.

There are refactoring tools in IDEs that can help with this.

We should aim to make methods no longer than 40 lines, but it's not a strict rule. Sometimes it's ok if they're longer.

tools/reviewer/aa/commands/DiffCommand.java
```diff
@@ -178,5 +184,38 @@ public boolean run(String[] args) {
...
+      fileUtils
+          .listContents(workspacePath)
+          .stream()
+          .map(path -> fileUtils.joinToAbsolutePath(workspacePath, path))
+          .filter(fileUtils::folderExists)
+          .forEach(
+              path -> {
+                String repoName = Paths.get(path).getFileName().toString();
+                GitRepo repo = this.gitRepoFactory.create(path);
+                String repoURL = repo.getRemoteURL();
```
oferb: I think you should get the `repoName` from the URL, like you do with `repoOwner`.
You could add a warning here if they're not the same.

tools/reviewer/aa/commands/DiffCommand.java
```diff
@@ -178,5 +184,38 @@ public boolean run(String[] args) {
     codeReviewBlockingStub.createDiff(request);
     return true;
   }
+
+  private void addGithubRepo(Diff.Builder diffBuilder) {
```
oferb: s/addGithubRepo/addGithubRepos

oferb: I think - since we're actually going to create PRs in these repos, we should only add repos which have changes in them. So how about you check for that instead?
You can add a method to `GitRepo` that checks if there are any changes from `master`. Something like `hasChages()`. It would check that `getCommits()` is not empty, or that `getUncommittedFiles()` is not empty.

PR #395, opened 2018-10-31
GitHub Sync: Add DiffReader

val-fed:


val-fed:
#396

tools/reviewer/job/sync/DiffReader.java
```diff
@@ -0,0 +1,39 @@
...
+
+package com.google.startupos.tools.reviewer.job.sync;
+
+import com.google.startupos.tools.reviewer.localserver.service.CodeReviewServiceGrpc;
+import com.google.startupos.tools.reviewer.localserver.service.Protos.Diff;
+import com.google.startupos.tools.reviewer.localserver.service.Protos.DiffRequest;
+import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
+
+public class DiffReader {
```
oferb: Maybe call this `ReviewerClient`?
I think you won't need a `ReviewerReader` and `ReviewerWriter` and can just use `ReviewerClient` for both reads and writes, because it's much simpler than the GitHub case - everything is proto and gRPC.

val-fed: Done.

PR #411, opened 2018-11-18
webapp - new file-changes threads

webdevelopland:
What's new:
- threads was updated a little bit to be more independent
- file-changes threads was changes to the threads (which was originally create for diff page)
- small design updates of file-changes page


tools/reviewer/webapp/src/app/dashboard/diff/diff-discussion/diff-discussion.component.ts
```diff
@@ -15,22 +15,17 @@ export class DiffDiscussionComponent {
...
-    this.threadStateService.openCommentChanges.subscribe(() => {
-      this.isExpanded = true;
-    });
+  // Changes "Expand" button to "Collapse", when at least one comment is expanded
+  commentExpanded(): void {
+    this.isExpanded = true;
   }
 
-  // Expand or collapse threads
+  // Expands or collapses all threads
```
oferb: Good job!

tools/reviewer/webapp/src/app/dashboard/diff/diff-discussion/thread-state.service.ts
```diff
@@ -0,0 +1,69 @@
...
+interface ThreadExpandedMap {
+  [id: string]: CommentExpandedMap;
+}
+
+interface FreezeMap {
+  [id: string]: boolean;
+}
+
+@Injectable()
+export class ThreadStateService {
```
oferb: Can you add some documentation on what type of state we need to save? What is freeze mode?

webdevelopland: There's explanation what freeze mode is inside thread component file. I don't know what do to best. Should I copy it everywhere, where freeze mode is used?
https://github.com/google/startup-os/blob/WD35/tools/reviewer/webapp/src/app/shared/thread/thread.component.ts#L14

tools/reviewer/webapp/src/app/dashboard/diff/diff-discussion/diff-discussion.component.ts
webdevelopland: :blush:

tools/reviewer/webapp/src/app/dashboard/diff/diff-discussion/thread-state.service.ts
oferb: Ok, that's fine.

PR #431, opened 2018-12-02
D347: Support Bazel 0.20.0 with new http_archive and http_file rules

oferb:
Note - this currently results in an error while building (`./build.sh`).
```
ERROR: /home/oferb/devel/base/ws/bazel_build_test/startup-os/tools/BUILD:64:12: in jars attribute of java_import rule //tools:bazel_deps_jar: '@bazel_deps//file:file' does not produce any java_import jars files (expected .jar)
```


vmax:
@oferb I fixed some `bazel`-related errors but now have trouble with Java versions

PR #432, opened 2018-12-03
GitHub Sync: Get patch using local Git

val-fed:


common/repo/GitRepo.java
```diff
@@ -461,6 +461,11 @@ public boolean hasChanges(String branch) {
     return (getCommits(branch).size() > 1) || (!getUncommittedFiles().isEmpty());
   }
 
+  @Override
+  public String getPatch(String commitOrBranch, String filename) {
```
oferb: s/commitOrBranch/referenceCommitOrBranch
According to the docs, this is for "changes you have in your working tree relative to the named <commit>".
https://git-scm.com/docs/git-diff

common/repo/Repo.java
```diff
@@ -81,5 +81,10 @@
   String getRemoteURL();
   /** Checks if there are commits added since master, or any uncommitted files */
   boolean hasChanges(String branch);
+  /**
+   * Gets patch(diff) between current file and the file of the base branch(usually "master"). It's
```
oferb: See previous comment - isn't this between the current file and the file in the passed `commitOrBranch`?

common/repo/repo.proto
```diff
@@ -44,6 +44,9 @@ message File {
   // Empty if referring to a local file
   string commit_id = 5;
   Action action = 6;
+  // Git patch(diff) between current file and the file of the base
```
oferb: Why do we need this new field?
If the repo is available to the sync job, we can always fetch it from the repo.

common/repo/Repo.java
val-fed: Done.

common/repo/Repo.java
```diff
@@ -81,5 +81,12 @@
   String getRemoteURL();
   /** Checks if there are commits added since master, or any uncommitted files */
   boolean hasChanges(String branch);
+  /**
+   * Gets patch(diff) between current file and the file in the passed `referenceCommitOrBranch`.
+   * It's possible to use branch name or commit ID. The response can have one and more diff hunks
+   */
+  String getPatch(String referenceCommitOrBranch, String filename);
+
+  String getTheLatestCommitIdOfBranch(String branch);
```
oferb: How about
`/getTheLatestCommitIdOfBranch/getMostRecentCommit`

tools/reviewer/job/sync/LineNumberConverter.java
```diff
@@ -164,7 +164,21 @@ private static int processDiffPatch(
       }
     } else {
       if (side.equals(Side.LEFT)) {
-        return lineNumberToPositionLeftSide.get(number);
+        if (lineNumberToPositionLeftSide.containsKey(number)) {
+          return lineNumberToPositionLeftSide.get(number);
+          // It isn't possible to convert Reviewer's line number to GitHub patch(diff)'s position if
+          // the patch doesn't contain this line.
+        } else {
+          // XXX 1)How to continue the sync without exception.
```
oferb: I would choose any line (maybe the closest, if that's possible), and add to the comment:
"Synced from line X"

common/repo/Repo.java
val-fed: Done.

tools/reviewer/job/sync/LineNumberConverter.java
val-fed: Added TODO for it.

PR #450, opened 2018-12-12
D363: Add tool to convert file to base64

oferb:


common/Base64Tool.java
```diff
@@ -0,0 +1,72 @@
...
+ * limitations under the License.
+ */
+
+package com.google.startupos.common;
+
+import com.google.startupos.common.CommonModule;
+import dagger.Component;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import com.google.startupos.common.FileUtils;
```
val-fed: unnecessary imports:
- import com.google.startupos.common.CommonModule;
- import com.google.startupos.common.FileUtils;

common/Base64Tool.java
```diff
@@ -0,0 +1,72 @@
...
+
+  private FileUtils fileUtils;
+
+  @Inject
+  Base64Tool(FileUtils fileUtils) {
+    this.fileUtils = fileUtils;
+  }
+
+  void run() throws IOException {
+    String fileContents = fileUtils.readFile(file.get());
```
val-fed: Maybe s / fileContents / fileContent ?
Also `encodedContents` and `decodedContents`.

common/Base64Tool.java
oferb: Done

common/Base64Tool.java
oferb: Actually, contents sounds better. Here's a related read:
https://jakubmarian.com/content-vs-contents-in-english/

PR #453, opened 2018-12-13
D368: Move aa_tool.sh logic into java

oferb:
Simplify aa by moving code completion, start server and killserver from bash to AaScriptHelper.java

This Diff also adds HttpUtils.java for HTTP-related code.

### DIFF
--- a/tools/reviewer/aa/AaScriptHelper.java
+++ b/tools/reviewer/aa/AaScriptHelper.java
@@ -24,17 +24,29 @@ import java.util.Map;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import javax.inject.Named;
+import java.io.IOException;
+import com.google.startupos.common.FileUtils;
+import com.google.startupos.common.HttpUtils;
+import java.io.File;
 
 /** Tool with some helpful functions for aa script (aa_tool.sh) */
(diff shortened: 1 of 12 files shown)


tools/reviewer/aa/AaScriptHelper.java
```diff
@@ -0,0 +1,173 @@
...
+  // bazelWorkspacePath is base/head/startup-os or base/ws/<some workspace>/startup-os
+  private boolean checkServer() {
+    try {
+      return "OK".equals(httpUtils.get("http://localhost:7000/health"));
+    } catch (Exception e) {
+      return false;
+    }
+  }
+
+  private void check(boolean predicate, String message) throws IllegalArgumentException {
```
vmax: @oferb strictly speaking, I think it's not `predicate`, it's the result of predicate evaluation. Citing [source](https://en.wikipedia.org/wiki/Predicate_(mathematical_logic))
>Predicate is commonly understood to be a Boolean-valued function P: X→ {true, false}, called the predicate on X. 

But I'm ok with it because function is small and concise 🙂

oferb: Good point. Changed to `predicateResult` :)

PR #456, opened 2018-12-16
SubmitterJob

vmax:
continues work in #448


common/firestore/MessageWithId.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package com.google.startupos.common.firestore;
+
+import com.google.auto.value.AutoValue;
+import com.google.protobuf.Message;
+
+@AutoValue
+public abstract class MessageWithId {
+  // A holder class for a proto message and an ID, typically the Firestore document ID.
```
oferb: This comment should be for the whole class.

tools/reviewer/job/README.md
```diff
@@ -26,3 +26,11 @@ from previous version, push it to Firestore.
 Pulls `CIRequest` from `/reviewer/ci/requests`, clones all the `Repo`s it
 contains, checks out relevant commits for each of them, runs `reviewer-ci.sh`
 and stores `CIResponse` in `/reviewer/ci/responses/`
+
+
+### SubmitterTask
```
oferb: Great!

tools/reviewer/job/impl/SubmitterTask.java
```diff
@@ -0,0 +1,187 @@
...
+
+        if (!fileUtils.folderExists("submitter")) {
+          fileUtils.mkdirs("submitter");
+        }
+
+        diffs
+            .stream()
+            // get Diffs in SUBMITTING state
+            .filter(diff -> diff.getStatus().equals(Diff.Status.SUBMITTING))
+            // for each of them, get latest CIResponse (get or pop?)
```
oferb: I think at this point, since we don't have a scheduler, let's get now, and remove after we succeed.

tools/reviewer/job/impl/SubmitterTask.java
```diff
@@ -0,0 +1,187 @@
...
+            .map(
+                diff -> {
+                  MessageWithId responseWithId =
+                      this.firestoreClient.getDocumentFromCollection(
```
oferb: So we get _a_ document, but will it be the most recent one?

tools/reviewer/job/impl/SubmitterTask.java
```diff
@@ -0,0 +1,187 @@
...
+                      }
+                    } catch (IOException e) {
+                      e.printStackTrace();
+                    }
+
+                    String branch = String.format("origin/D%d", diff.getId());
+                    gitRepo.switchBranch(branch);
+
+                    newRequestBuilder =
+                        newRequestBuilder.addTarget(
```
oferb: Do we need to assign this back to `newRequestBuilder`?

tools/reviewer/job/impl/SubmitterTask.java
```diff
@@ -0,0 +1,187 @@
...
+                        gitRepos.stream().allMatch(repo -> repo.push("master"));
+                    if (allPushesSuccessful) {
+                      diff = diff.toBuilder().setStatus(Diff.Status.SUBMITTED).build();
+                      this.firestoreClient.createProtoDocument(
+                          DIFF_PATH, String.valueOf(diff.getId()), diff);
+                    } else {
+                      throw new RuntimeException("Not all pushes were successful");
+                    }
+                  } else {
+                    if (!ciSubmittedDiffIds.contains(diff.getId())) {
```
oferb: Why don't we hash the `newRequestBuilder.build()` object?
We might want to run a different request for the same diff.

tools/reviewer/job/impl/SubmitterTask.java
```diff
@@ -0,0 +1,187 @@
...
+                                .setRepo(repo)
+                                .setCommitId(gitRepo.getHeadCommitId())
+                                .build());
+
+                    if (gitRepo.getHeadCommitId().equals(target.getCommitId())) {
+                      // latest result for CI matches last commit on the branch
+
+                      gitRepo.switchBranch("master");
+                      boolean mergingResult = gitRepo.merge("D%d", true);
+                      if (!mergingResult || gitRepo.hasChanges("master")) {
```
oferb: Do we have a place to store the failed result? For example - `Merge failed` or `Commit out of date`.

tools/reviewer/job/impl/SubmitterTask.java
vmax: @oferb 
>We might want to run a different request for the same diff.

why is that? I think `SubmitterTask` only submits new requests in this one specific case. Generally, we rely on `aa` doing it.

tools/reviewer/job/impl/SubmitterTask.java
vmax: @oferb Nope. what is the best way to handle it?

tools/reviewer/job/impl/SubmitterTask.java
vmax: @oferb we probably don't as it compiles to `this = this` (`this` being `newRequestBuilder`) which doesn't make much sense

tools/reviewer/job/impl/SubmitterTask.java
vmax: @oferb I don't think CI responses should be removed at all. Shouldn't we track CI history for every diff?

common/firestore/MessageWithId.java
vmax: @oferb ae17bcd6192338e64122e84142d2a2c4c2d42604

tools/reviewer/job/impl/SubmitterTask.java
vmax: @oferb ae17bcd6192338e64122e84142d2a2c4c2d42604

tools/reviewer/job/impl/SubmitterTask.java
oferb: Ok, you're right.

tools/reviewer/job/impl/SubmitterTask.java
vmax: @oferb makes sense!

tools/reviewer/job/impl/SubmitterTask.java
vmax: @oferb do you mean we should add `field` to proto or new proto message class?

oferb: I don't mind. Could be a message, or 2 separate fields in an existing message.

vmax: @oferb I think it makes sense to have `SubmitterMergeResult` as new proto message which will be stored in `/reviewer/submitter/<diffId>/history/`. I'll create a message & add a `TODO` to store result

PR #460, opened 2018-12-19
webapp - open in new tab

webdevelopland:
Ability to open link in new tab


tools/reviewer/webapp/src/app/dashboard/diff/diff.service.ts
```diff
@@ -1,15 +1,31 @@
...
-      'diff/' + diffId + '/' + file.getFilenameWithRepo(),
-    ], { queryParams: { right: commitId } });
+  openFile(
+    isNewTab: boolean,
+    file: File,
+    diffId: number,
+    commitId?: string,
+  ): void {
+    const url: string = 'diff/' + diffId + '/' + file.getFilenameWithRepo();
+    const queryParams: Params = { right: commitId };
```
oferb: Looks like you only need `const queryParams: Params = { right: commitId };` in the `else`.

PR #461, opened 2018-12-19
GitHub Sync: Add sync logic

val-fed:


tools/reviewer/job/sync/full_github_pr.proto
```diff
@@ -52,6 +52,9 @@ message ReviewComment {
   string updated_at = 13;
   Link _links = 14;
   string reviewer_thread_id = 15;
+  string reviewer_comment_id = 16;
+  bool is_outside_diff_comment = 17;
+  int32 reviewer_line_number = 18;
```
oferb: If you have the Reviewer commend Id, why do you need the `reviewer_line_number`? Isn't that information already stored there?

val-fed: Actually, the `reviewer_line_number` field is needed for a case when GitHub patch doesn't contain the line on we want to add the comment. 
It happens here: 
https://github.com/google/startup-os/blob/fa9da87e2270a3677dee8599eebc2f9fb78cf9c2/tools/reviewer/job/sync/GithubWriter.java#L234
`reviewer_line_number` is a number of the line where the comment was added. 
`reviewer_comment_id` and `reviewer_thread_id` allow to identify exactly the comment in Reviewer. It looks like `QJzjy`.

oferb: Ok, cool.

PR #463, opened 2018-12-20
D376: Refactor reviewer job code

oferb:
* Move all task-related code into one folder
* Split into BUILD targets
* Make TaskExecutor accept a list of tasks.


tools/reviewer/job/ReviewerJob.java
```diff
@@ -59,19 +63,16 @@
...
-  public static Flag<String> serviceAccountJson = Flag.create("");
-
-  @FlagDesc(name = "repo_url", description = "", required = false)
-  public static Flag<String> repoUrl = Flag.create("");
-
   @Inject
   public ReviewerJob(
-      TaskExecutor taskExecutor,
       InitCommand initCommand,
       Lazy<LocalServer> lazyLocalServer,
```
vmax: @oferb why are `LocalServer` and `HeadUpdater` `Lazy<>`?

tools/reviewer/job/tasks/BUILD
```diff
@@ -0,0 +1,179 @@
...
+
+java_library(
+    name = "task_executor",
+    srcs = [
+        "TaskExecutor.java",
+    ],
+    deps = [
+        ":reviewer_metadata_updater_task",
+        ":task_interface",
+        "//common",
```
vmax: @oferb it seems that this part of `deps` is repeated pretty often across this `BUILD` file. Do you think it's worth separating out into a variable or a separate `java_library` with `exports`?

oferb: Actually, most of it is not needed, but it will get removed anyway once we have automated BUILD files, so I'm being `Lazy<>` :-P

tools/reviewer/job/ReviewerJob.java
oferb: Good point. Forgot to document it. Added this:
```
// These are lazy because constructing them requires us to be in a base folder.
// That base folder is only set up in run() using initCommand. We also assumme we're inside it.
```

PR #465, opened 2018-12-21
webapp - CI status

webdevelopland:
The PR adds CI UI to diff page


tools/reviewer/reviewer.proto
```diff
@@ -25,9 +25,18 @@ message CIRequest {
...
+      RUNNING = 3;
+      OUTDATED = 4;
+    }
     CIRequest.Target target = 1;
-    bool success = 2;
-    string log = 3;
+    Status status = 2;
+    // Log weighs too much
+    // TODO: move log somewhere else
+    // string log = 3;
```
oferb: It's ok, you can leave the `log` here, we just won't fill it up.
We;ll use the same proto and store the log somewhere else in Firestore inside it.

PR #466, opened 2018-12-22
Build File Generator: Add Java class Analyzer

val-fed:


tools/build_file_generator/JavaClassAnalyzer.java
```diff
@@ -0,0 +1,97 @@
...
+
+import com.google.startupos.tools.buildfilegenerator.Protos.Import;
+import com.google.startupos.tools.buildfilegenerator.Protos.JavaClass;
+
+public class JavaClassAnalyzer {
+  public static JavaClass getJavaClass(String fileContent) {
+    JavaClass.Builder result = JavaClass.newBuilder();
+
+    List<String> importLines = getImportLines(fileContent);
+    importLines.forEach(line -> result.addImport(getImport(line)));
```
oferb: I think you can do:
`getImportLines(fileContent).forEach`
Unless it simplifies a complex expression, I tend to prefer not to introduce new local variables.

tools/build_file_generator/JavaClassAnalyzer.java
```diff
@@ -0,0 +1,97 @@
...
+    result.setIsTestClass(isTestClass(fileContent));
+    result.setHasMainMethod(hasMainMethod(fileContent));
+
+    return result.build();
+  }
+
+  private static List<String> getImportLines(String fileContent) {
+    return Arrays.stream(fileContent.split(System.lineSeparator()))
+        .filter(
+            line ->
```
oferb: Maybe you can trim the string, to handle the case it starts with spaces?

tools/build_file_generator/JavaClassAnalyzer.java
```diff
@@ -0,0 +1,97 @@
...
+
+  private static Import getImport(String importLine) {
+    Import.Builder importBuilder = Import.newBuilder();
+
+    boolean isStaticImport = importLine.split(" ")[1].equals("static");
+
+    String lastImportPart;
+    List<String> importParts;
+    if (isStaticImport) {
+      String importBody = importLine.split(" ")[2].replace(";", "");
```
oferb: Maybe you can do some pre-processing so that there's more reuse between the static and non-static cases?

tools/build_file_generator/JavaClassAnalyzer.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+    String rootDir = importParts.get(0);
+
+    importBuilder.setRootDir(rootDir).setStandardJavaPackage(isStandardJavaPackage(rootDir));
+    return importBuilder.build();
+  }
+
+  private static boolean isStandardJavaPackage(String rootImportDir) {
+    // XXX: Is it needed to create dependencies for classes starts with "javax"?
+    List<String> standardJavaPackages = Arrays.asList("java", "javax");
```
oferb: I don't think so.. but it's good to have a tool that maps all deps. We might use it later for something else.

tools/build_file_generator/JavaClassAnalyzer.java
```diff
@@ -0,0 +1,97 @@
...
+    return standardJavaPackages.contains(rootImportDir);
+  }
+
+  private static boolean isTestClass(String fileContent) {
```
oferb: Nice heuristic. Should work.. :)
Also for `hasMainMethod`

tools/build_file_generator/build_file_generator.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+package com.google.startupos.tools.buildfilegenerator;
+
+option java_package = "com.google.startupos.tools.buildfilegenerator";
+option java_outer_classname = "Protos";
+
+message Import {
+  string root_dir = 1;
```
oferb: Can you explain what these fields are with an example?

tools/build_file_generator/build_file_generator.proto
```diff
@@ -0,0 +1,19 @@
...
+  repeated string subdir = 2;
+  string imported_class_name = 3;
+  bool whole_package_import = 4;
+  bool standard_java_package = 5;
+}
+
+message JavaClass {
+  repeated Import import = 1;
```
oferb: We can get the class name from the filename, right?
We'll also need the package, which you can parse as well.

tools/build_file_generator/build_file_generator.proto
val-fed: Actually, `root_dir` field doesn't make sense, deleted it.
For example, this class:

```
package com.google.startupos.tools.buildfilegenerator;

import com.google.startupos.common.flags.Flag;
import com.google.startupos.common.flags.FlagDesc;
import com.google.startupos.common.flags.Flags;

import java.util.Arrays;
import java.util.List;

public class JavaClass {
  @FlagDesc(name = "file_path", description = "Path to file")
  static Flag<String> filePath = Flag.create("~/JavaClass.java");

  public static void main(String[] args) {
    Flags.parseCurrentPackage(args);

    List<String> classes = Arrays.asList(filePath.get());
  }
}
```

will be parsed:

```
classname: "JavaClass"
package: "com.google.startupos.tools.buildfilegenerator"
import {
  import_dir: "com"
  import_dir: "google"
  import_dir: "startupos"
  import_dir: "common"
  import_dir: "flags"
  imported_class_name: "Flag"
}
import {
  import_dir: "com"
  import_dir: "google"
  import_dir: "startupos"
  import_dir: "common"
  import_dir: "flags"
  imported_class_name: "FlagDesc"
}
import {
  import_dir: "com"
  import_dir: "google"
  import_dir: "startupos"
  import_dir: "common"
  import_dir: "flags"
  imported_class_name: "Flags"
}
import {
  import_dir: "java"
  import_dir: "util"
  imported_class_name: "Arrays"
  standard_java_package: true
}
import {
  import_dir: "java"
  import_dir: "util"
  imported_class_name: "List"
  standard_java_package: true
}
has_main_method: true
```

val-fed: Added the comments to the proto file.

tools/build_file_generator/JavaClassAnalyzer.java
val-fed: Done.

tools/build_file_generator/JavaClassAnalyzer.java
val-fed: Done.

tools/build_file_generator/JavaClassAnalyzer.java
val-fed: Done.

tools/build_file_generator/build_file_generator.proto
oferb: I don't think we need the split to `import_dir`.
How about we use this structure:
```

File: tools/build_file_generator/JavaClassAnalyzer.java [hunk 8]
```diff
@@ -0,0 +1,134 @@
...
+
+  private static List<String> getImportLines(String fileContent) {
+    return getLinesStartWithKeyword(fileContent, "import", ".");
+  }
+
+  private static List<String> getLinesStartWithKeyword(
+      String fileContent, String keyword, String lineShouldContain) {
+    return Arrays.stream(fileContent.split(System.lineSeparator()))
+        .map(String::trim)
+        .collect(Collectors.toList())
```
oferb: Do we need to collect and then stream again? What happens if we don't?

val-fed: Yes, you are right. It isn't necessarily. 
The redundant call deleted.

vmax:
@val-fed @oferb see how I broke `cla/google` but fixed `circleci/build` — net zero 😄

tools/build_file_generator/build_file_generator.proto
val-fed: The code is changed to work with the new structure.

tools/build_file_generator/JavaClassAnalyzer.java
```diff
@@ -0,0 +1,108 @@
...
+    String[] importLineParts = importLine.replace("import ", "").replace(";", "").split(" ");
+
+    boolean isStaticImport =
+        importLineParts.length > 1 && importLineParts[0].trim().equals("static");
+    result.setIsStatic(isStaticImport);
+
+    String[] importBodyParts =
+        isStaticImport ? importLineParts[1].split("\\.") : importLineParts[0].split("\\.");
+    if (importBodyParts.length < 3) {
+      throw new IllegalArgumentException("Import is too broad: " + importLine);
```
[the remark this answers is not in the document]
val-fed: I mean such kind of import: `import java.*;`

PR #467, opened 2018-12-22
Fix TODOs in aa_tool.sh; implement set_STARTUP_OS_REPO, bazel_run

vmax:


tools/reviewer/aa/aa_tool.sh
```diff
@@ -10,16 +10,45 @@
...
+    fi
+}
+
+function bazel_run() {
+    local target=$1
+    shift;
+    local force_compile=$1
+    shift;
+    local args="$@"
+    local binary_for_target=$(echo "$target" | sed -e 's|//||g; s|:|/|g; s|^|bazel-bin/|g')
```
oferb: Maybe place `s|^|bazel-bin/|g` first, so that the replaces are in order?
Also please explain what the `sed` command does.

tools/reviewer/aa/aa_tool.sh
```diff
@@ -10,16 +10,45 @@
...
+
+    set_STARTUP_OS_REPO
+    pushd $STARTUP_OS_REPO
+
+    if [[ ! -f ${binary_for_target} ]] || [[ "${force_compile}" -eq 1 ]]; then
+        bazel build ${target} &> /dev/null
```
oferb: We didn't `cd` into `STARTUP_OS_REPO` right?
Does `bazel` run it from there then?

vmax: @oferb 

>We didn't cd into STARTUP_OS_REPO right?
we did in line 38 😉 

`pushd $STARTUP_OS_REPO`

tools/reviewer/aa/aa_tool.sh
vmax: @oferb
>Maybe place s|^|bazel-bin/|g first, so that the replaces are in order?

we could do that

>Also please explain what the sed command does.

`sed` replaces bazel target name with name of the binary `bazel build` would produce
by doing the following (`sed` commands are split by `;`):
1. replace `//` with nothing
2. replace `:` with `/`
3. add `bazel-bin/` to the beginning

tools/reviewer/aa/aa_tool.sh
oferb: Right.. ;)

PR #476, opened 2018-12-24
Build code several times if on CI

vmax:


vmax:
@oferb good point! idk what happened with my logic when I wrote it 😄

PR #500, opened 2019-01-12
Clone all repos specified in global_registry config

vmax:


tools/reviewer/aa/commands/InitCommand.java
```diff
@@ -52,6 +57,14 @@ public InitCommand(FileUtils fileUtils, GitRepoFactory gitRepoFactory) {
     this.gitRepoFactory = gitRepoFactory;
   }
 
+  private void cloneRepoIntoHead(String dirName, String repoUrl) {
+    String startupOsPath = fileUtils.joinToAbsolutePath(basePath.get(), "head", dirName);
+    System.out.println("Cloning StartupOS into " + startupOsPath);
```
oferb: It's now not necessarily StartupOS, so something like s/StartupOS/repoUrl.

tools/reviewer/aa/commands/InitCommand.java
```diff
@@ -42,6 +44,9 @@
   public static Flag<String> startuposRepo =
       Flag.create("https://github.com/google/startup-os.git");
 
+  @FlagDesc(name = "global_registry_config", description = "Path to global_registry.prototxt")
```
oferb: I think we can take the global registry directly from the repo itself, so no need for a flag.

vmax: @oferb fixed in https://github.com/google/startup-os/pull/500/commits/e0915c92017ded4188e955585e1109bab2bd7176

now the path is fixed at `tools/reviewer/global_registry.prototxt`. However, I'm not sure relative to _which_ repo it should be appended so pls clarify

tools/reviewer/aa/commands/InitCommand.java
vmax: @oferb fixed in https://github.com/google/startup-os/pull/500/commits/e0915c92017ded4188e955585e1109bab2bd7176

tools/reviewer/aa/commands/InitCommand.java
oferb: The current repo.
The way it would work, is you would clone `startup-os`, and then run `aa init`.
So you would use whatever config is currently in `master`.

tools/reviewer/aa/commands/InitCommand.java
```diff
@@ -73,15 +85,27 @@ public boolean run(String basePath, String startuposRepo) {
...
-        System.out.println("Cloning StartupOS into " + startupOsPath);
-        GitRepo repo = this.gitRepoFactory.create(startupOsPath);
-        repo.cloneRepo(startuposRepo, startupOsPath);
-        System.out.println("Completed Cloning");
+        cloneRepoIntoHead("startup-os", startuposRepo);
       } else {
         System.out.println("Warning: StartupOS repo url is empty. Cloning skipped.");
       }
 
+      if (!GLOBAL_REGISTRY_CONFIG.isEmpty()) {
```
oferb: No need to check if `GLOBAL_REGISTRY_CONFIG` is not empty.

PR #515, opened 2019-01-23
Build File Generator: Generate BUILD files for whole repo

val-fed:
- Target names are created in the lower underscore format
- BuildFileGenerator generate BUILD files for the whole repo

### DIFF
--- a/tools/build_file_generator/BuildFileGenerator.java
+++ b/tools/build_file_generator/BuildFileGenerator.java
@@ -36,7 +36,9 @@ import com.google.startupos.tools.build_file_generator.Protos.ThirdPartyDeps;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
 
@@ -54,6 +56,7 @@ public class BuildFileGenerator {
   private ProtoFileAnalyzer protoFileAnalyzer;
   private JavaClassAnalyzer javaClassAnalyzer;
   private ThirdPartyDeps thirdPartyDeps;
+  private List<String> pathsToGenerateBuildFiles;
 
   @Inject
   public BuildFileGenerator(
@@ -64,6 +67,7 @@ public class BuildFileGenerator {
     this.protoFileAnalyzer = protoFileAnalyzer;
     this.javaClassAnalyzer = javaClassAnalyzer;
     thirdPartyDeps = getThirdPartyDeps();
+    pathsToGenerateBuildFiles = getPathsToGenerateBuildFiles();
   }
 
   private ThirdPartyDeps getThirdPartyDeps() {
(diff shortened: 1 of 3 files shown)


tools/build_file_generator/BuildFileGenerator.java
```diff
@@ -71,52 +76,78 @@ private ThirdPartyDeps getThirdPartyDeps() {
             THIRD_PARTY_ZIP_PATH, PROTOTXT_FILENAME_INSIDE_ZIP, ThirdPartyDeps.newBuilder());
   }
 
-  BuildFile generateBuildFile(String packagePath) throws IOException {
-    BuildFile.Builder result = BuildFile.newBuilder();
+  private List<String> getPathsToGenerateBuildFiles() {
```
oferb: How about `s/getPathsToGenerateBuildFiles/getCodeFilePaths`

tools/build_file_generator/BuildFileGenerator.java
```diff
@@ -63,6 +67,7 @@ public BuildFileGenerator(
     this.protoFileAnalyzer = protoFileAnalyzer;
     this.javaClassAnalyzer = javaClassAnalyzer;
     thirdPartyDeps = getThirdPartyDeps();
+    pathsToGenerateBuildFiles = getPathsToGenerateBuildFiles();
```
oferb: Why do this in the constructor and not in `generateBuildFiles()`?
Actually, also for `thirdPartyDeps = getThirdPartyDeps();`.

tools/build_file_generator/BuildFileGenerator.java
```diff
@@ -71,52 +76,78 @@ private ThirdPartyDeps getThirdPartyDeps() {
...
+          }
+        }
+      }
+    } catch (IOException e) {
+      e.printStackTrace();
+    }
+    return result;
+  }
+
+  Map<String, BuildFile> generateBuildFiles() throws IOException {
```
oferb: Please document the result, what is `String`?

oferb: I think it was actually nice to have a method for `generateBuildFile()` that gets called by `generateBuildFiles()`, especially since it's already quite a large method, so that would sort of separate the "single element" logic from the 'apply it to a list" logic.

val-fed: Done.

val-fed: Done.

tools/build_file_generator/BuildFileGenerator.java
val-fed: Done.

tools/build_file_generator/BuildFileGenerator.java
val-fed: Done.

tools/build_file_generator/BuildFileGenerator.java
```diff
@@ -96,58 +92,66 @@ private ThirdPartyDeps getThirdPartyDeps() {
     return result;
   }
 
+  /* Returns Map<String, BuildFile> where
+  String kay is an absolute path to a folder where BUILD file should be created,
```
oferb: s/kay/key

val-fed: Done.

PR #519, opened 2019-01-30
Web login - simple html + js version

webdevelopland:


tools/reviewer/local_server/web_login_js/src/style.css
```diff
@@ -0,0 +1,20 @@
+.google {
```
oferb: Can we do it reasonably without the `login-with-google.png` and `favicon.ico`?

tools/reviewer/local_server/web_login_js/src/index.html
```diff
@@ -0,0 +1,26 @@
...
+  <div class="google">
+    <div class="gbutton" id="login">
+      <img onclick="app.login()" src="/login-with-google.png">
+    </div>
+    <div class="gbutton" id="logout">
+      <button onclick="app.logout()">
+        Log Out
+      </button>
+    </div>
+    <div class="chrome" id="chrome">For best support please open chrome</div>
```
oferb: :-)

oferb: `s/chrome/Chrome`

tools/reviewer/local_server/web_login_js/README.md
```diff
@@ -0,0 +1,14 @@
+# Firestore web login for StartupOS local server
+
+### Setting
+Port and config: `./src/config.js`
+
+### Development
+```
+npm start
```
oferb: If this is just plain HTML+js, why do we need npm? Please document.

webdevelopland: There was a small development `expressjs` server, which launched the HTML+js code locally.
`expressjs` require npm.
But we can run localserver from outside, it doesn't have to be part of the project. So I removed it.

tools/reviewer/local_server/web_login_js/firebase.json
```diff
@@ -0,0 +1,27 @@
+{
```
oferb: Cool! Maybe we can further simplify by putting all the js inside the html file?
That way you can actually open the file in the browser and it will work, you don't need a server.
Or is that frowned upon? :-)

webdevelopland: Good idea :)

webdevelopland: Looks like we can't work with firebase, when all the code is launched from a single file.
Error:
```
This operation is not supported in the environment…chrome-extension and web storage must be enabled.
```

webdevelopland: But the file work, if it's launched from server.
So I can put everything in one file, but there's no much sense in it.

oferb: Ok, let's still keep it in one file. I think it will be easier to host it that way.

PR #530, opened 2019-02-15
Add support for easy bazel version upgrade

Jdban:
With this change, when the `.bazelversion` file at the root of the repository changes, the new version is downloaded and any old versions are deleted


oferb:
What if you have multiple repositories, each with a different version?
Or you're working on 2 features - one branching off `master` before a version bump, and one after?
Every time you switch features or repositories and build, it will download the binary.

I agree that the old version should be cleaned. Perhaps we could count the old versions, and if there's more than 3, print a warning to the user to manually delete them?

PR #545, opened 2019-03-08
Make getting third party deps independent 

val-fed:


tools/build_file_generator/third_party_deps/create_third_party_deps_zip.sh
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
```
oferb: Why did we rewrite the java code in bash?
Couldn't we run the java code on the same input as the bash code?

tools/build_file_generator/BUILD
```diff
@@ -129,73 +114,3 @@ checkstyle_test(
     name = "build_file_generator_tool-checkstyle",
     target = ":build_file_generator_tool",
 )
-
```
oferb: To run this in other repos (e.g `hasadna`), you can still use the same code (ThirdPartyDepsTool), it's just the BUILD file that needs to be define in `hasadna`, no?

Or do you mean this would allow `hasadna` to reference code in `startup-os` more easily?

PR #553, opened 2019-03-19
Add grpc services processing, add auto-generated BUILD files

val-fed:
Add auto-generated BUILD files:
- `tools/reviewer/job/sync/tests/BUILD`
- `tools/reviewer/local_server/BUILD`

### DIFF
--- a/tools/build_file_generator/BuildFileGenerator.java
+++ b/tools/build_file_generator/BuildFileGenerator.java
@@ -368,33 +368,29 @@ public class BuildFileGenerator {
       if (target.equals("//common/repo:git_repo_factory")) {
         target = "//common/repo:repo";
       }
-      if (target.contains("code_review_service_grpc")) {
-        target = target.replace("_service_grpc", "_java_grpc");
-      }
-      if (target.contains("auth_service_grpc")) {
-        target = target.replace("_grpc", "_java_grpc");
-      }
-      if (target.endsWith("_java_proto")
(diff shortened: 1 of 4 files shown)


tools/build_file_generator/BuildFileGenerator.java
```diff
@@ -371,6 +371,9 @@ private CheckstyleTestExtension getCheckstyleTest(String target) {
       if (target.contains("code_review_service_grpc")) {
         target = target.replace("_service_grpc", "_java_grpc");
       }
+      if (target.contains("auth_service_grpc")) {
```
oferb: Why can't we also use:
`target = target.replace("_service_grpc", "_java_grpc");`
Instead of:
`target = target.replace("_grpc", "_java_grpc");`
?

val-fed: The problem is in names. `auth_service.proto` already involve `service` in the name. Name of service is AuthService. (`proto file name = AuthService`). 
The second case has `code_review.proto`. Name of service: CodeReviewService (`proto file name != CodeReviewService`)

oferb: Right. But when we parse proto files, we know all the data:
- proto filename
- java package
- service name

So maybe we can use that information to do this in a general way?

val-fed: Added the code to do this in a general way.

PR #554, opened 2019-03-23
Make third party deps tool available to use outside startup-os

val-fed:


tools/build_file_generator/third_party_deps/ThirdPartyDepsAnalyzer.java
```diff
@@ -0,0 +1,152 @@
+/*
```
oferb: Did you change this code, or just move it?

tools/build_file_generator/third_party_deps/third_party_deps_updater.sh
```diff
@@ -0,0 +1,184 @@
+#!/usr/bin/env bash
```
oferb: How did we do this before this script?

tools/build_file_generator/third_party_deps/ThirdPartyDepsAnalyzer.java
val-fed: The code has lots of changes. I created draft PR to see all them - https://github.com/google/startup-os/pull/557/files

tools/build_file_generator/third_party_deps/third_party_deps_updater.sh
val-fed: In the beginning I used `bazel query 'attr("visibility", "//visibility:public", //third_party/...)'` command and fill in deps in the BUILD file manually. 
We didn't have automation way for updating the BUILD file after adding/removing a third-party dependency in the project.

tools/build_file_generator/third_party_deps/third_party_deps_updater.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# A script for updating third party dependencies inside third_party_deps_tool java_binary
+
+readonly REPO_ABS_PATH=$(git rev-parse --show-toplevel)
+# TODO: Make the getting of the path independent of the project
+readonly BUILD_FILE_ABS_PATH="$REPO_ABS_PATH/tools/build_file_generator/third_party_deps/BUILD"
+
+# Getting public third-party target names
+third_party_folder_path="$REPO_ABS_PATH/third_party"
```
oferb: Why is this not `readonly` like the others, and styles as a constant?

tools/build_file_generator/third_party_deps/third_party_deps_updater.sh
```diff
@@ -0,0 +1,24 @@
...
+declare -a actual_third_party_targets
```
vmax: @val-fed 
```suggestion
actual_third_party_targets=()
```

tools/build_file_generator/third_party_deps/third_party_deps_updater.sh
```diff
@@ -0,0 +1,24 @@
...
+for entry in $(bazel query 'attr("visibility", "//visibility:public", //third_party/...)'); do
+  actual_third_party_targets=("${actual_third_party_targets[@]}" "$entry")
```
vmax: @val-fed 
```suggestion
  actual_third_party_targets+=("$entry")
```

tools/build_file_generator/third_party_deps/third_party_deps_updater.sh
```diff
@@ -0,0 +1,24 @@
...
+done
+
+# Deleting old targets
+sed -i '/# %START%/,/# %END%/{//!d}' ${BUILD_FILE_ABS_PATH}
```
vmax: @val-fed
There are couple of concerns here:
- `sed`, to be honest, is pretty cryptic to read. Maybe explain in detail what `sed` is meant to do here in terms of text manipulation ("delete text between %START and %END, while preserving the markers"). This way we'll be able to think of a better way to achieve the same goal.
- `-i <command> <file>` will only work on GNU sed. To make it portable you should do `sed -i.bak -e command file && rm file.bak`. See [here](https://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux).
- Also, your version doesn't work on BSD sed 😉. Please add a semicolon at the end so it does not fail with `extra characters at the end of d command`: 
```suggestion
sed -i '/# %START%/,/# %END%/{//!d;}' ${BUILD_FILE_ABS_PATH}
```

tools/build_file_generator/third_party_deps/third_party_deps_updater.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# Adding new targets
+for target_to_add in "${actual_third_party_targets[@]}"; do
+  sed -i "/# %START%/a \"${target_to_add}\"," ${BUILD_FILE_ABS_PATH}
```
vmax: @val-fed please see the comment on `sed` portability above

tools/build_file_generator/third_party_deps/third_party_deps_updater.sh
val-fed: This variable is never used, I forgot to remove it. 
Removed this line.

PR #560, opened 2019-03-30
Fix ThirdPartyDepsAnalyzer

val-fed:


tools/build_file_generator/third_party_deps/BUILD
```diff
@@ -21,42 +21,41 @@ checkstyle_test(
     target = ":third_party_deps_analyzer",
 )
 
-# '%START%' and '%END%' are markers for 'third_party_deps_updater.sh' script.
-# Don't delete it while 'third_party_deps_updater.sh' script is used.
+# '%START%' and '%END%' are markers for 'third_party_deps_updater.sh' script and ThirdPartyDepsAnalyzer class.
+# Don't delete it while 'third_party_deps_updater.sh' script and ThirdPartyDepsAnalyzer class are used.
 THIRD_PARTY_DEPS_GROUP = [
     # %START%
```
oferb: Maybe we should rename to `THIRD_PARTY_DEPS_START`, and similarly for `END`?

tools/build_file_generator/third_party_deps/BUILD
```diff
@@ -21,42 +21,41 @@ checkstyle_test(
...
     "//third_party/maven/com/google/dagger:dagger_compiler",
     "//third_party/maven/com/google/errorprone:error_prone_ant",
     "//third_party/maven/com/google/firebase:firebase_admin",
-    "//third_party/maven/com/google/flogger",
+    "//third_party/maven/com/google/flogger:flogger",
     "//third_party/maven/com/google/flogger:flogger_system_backend",
     "//third_party/maven/com/google/googlejavaformat:google_java_format",
-    "//third_party/maven/com/google/guava",
-    "//third_party/maven/com/google/jimfs",
+    "//third_party/maven/com/google/guava:guava",
```
oferb: There's no need to repeat the target if the folder name is the same:
`//third_party/maven/com/google/guava:guava`
Is the same as:
`//third_party/maven/com/google/guava`

Is it easy to shorten them?

tools/build_file_generator/third_party_deps/BUILD
val-fed: Done.

tools/build_file_generator/third_party_deps/BUILD
val-fed: Added logic to shorten it to the script.

PR #561, opened 2019-03-31
Restructure CI config

vmax:
- Upgrade `bazel` to `0.24.0`
- Restructure `.circleci/config.yml` to use `machine` executor. This removes memory limitations (see bazelbuild/bazel#3645) and removes the need to upgrade `libstdc++` (Ubuntu's bundled on is up-to-date enough)
- Disable building code multiple times on CircleCI


tools/build_or_test.sh
```diff
@@ -35,20 +33,5 @@ if [[ $1 != "build" && $1 != "test" ]]; then
...
-  for i in $(seq 1 ${CIRCLECI_MAX_ATTEMPTS}); do
-    echo "$RED""[Attempt $i/${CIRCLECI_MAX_ATTEMPTS}]: building$RESET"
-    bazel_build $1
-    if [[ $? -eq 0 ]]; then
-      echo "$GREEN""[Attempt $i/${CIRCLECI_MAX_ATTEMPTS}]: successful$RESET"
-      exit 0
-    fi
-  done
-
-  echo "$RED""[Attempts exhausted]: Seems it's a problem with your code and not a CircleCI flake.$RESET"
```
oferb: How come we don't need this anymore?

.circleci/config.yml
```diff
@@ -1,71 +1,56 @@
...
-      - restore_cache:
-          key: "v7-bazel_cache"
-
-      # Fixes Debian upstream repos: https://discuss.circleci.com/t/failed-to-fetch-jessie-updates/29246
-      - run: |
-          sudo rm /etc/apt/sources.list
-          echo "deb http://archive.debian.org/debian/ jessie-backports main" | sudo tee -a /etc/apt/sources.list
-          echo "deb-src http://archive.debian.org/debian/ jessie-backports main" | sudo tee -a /etc/apt/sources.list
-          echo "Acquire::Check-Valid-Until false;" | sudo tee -a /etc/apt/apt.conf.d/10-nocheckvalid
-          echo 'Package: *\nPin: origin "archive.debian.org"\nPin-Priority: 500' | sudo tee -a /etc/apt/preferences.d/10-archive-pin
```
oferb: We don't need `yapf3`? How do we format python code?

.circleci/config.yml
```diff
@@ -1,71 +1,56 @@
...
 
-# These jobs will run in parallel, and report separate statuses to GitHub PRs
+      - save_cache:
+          key: "v8-bazel_cache"
+          paths:
+            - "/home/circleci/.cache/bazel/"
 jobs:
   lint:
-    <<: *job_defaults
+    working_directory: ~/ng/
```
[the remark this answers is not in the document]
vmax: @oferb frankly speaking, because of my laziness to update `workspace_patcher`: https://github.com/google/startup-os/blob/d98b1eddc4989ef4f11d27d4914a4565df8cc9c6/tools/deps/WorkspacePatcher.java#L40

We can give it more generic name (i.e. `project`) and replace it both in `working_directory` and the line I linked above

.circleci/config.yml
```diff
@@ -89,19 +73,9 @@ jobs:
           command: cd /home/circleci/hasadna/ && ./compile.sh build
```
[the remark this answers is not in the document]
vmax: @oferb it matches the behaviour we currently have: https://github.com/google/startup-os/blob/d98b1eddc4989ef4f11d27d4914a4565df8cc9c6/.circleci/config.yml#L87-L91

Initially we did it because `bazel test` with no tests exits with [exit code 4](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/util/ExitCode.java#L48). Currently, we could do it, but there's only `checkstyle` tests which cannot really be broken by `@startup_os` — thus, we are only building.

.circleci/config.yml
vmax: @oferb we don't have Python code in `@startup_os`/`@hasadna`/`@devenv` — at least as far as I've checked — so I decided to simplify config to not install `yapf`

tools/build_or_test.sh
vmax: @oferb problems we are having with `bazel` is that in Docker container memory ceiling is not calculated correctly (see the issue I linked in description and [this comment](https://github.com/bazelbuild/bazel/issues/3645#issuecomment-331672268) specifically). 
That way, we re-ran `bazel` until it finally succeeded in hope that it would consume less memory with each run (due to caching). It is no longer needed because we'll be running on a real virtual machine instead of Docker — which means `bazel` is able to calculate memory ceiling correctly and not allocate more than available.

.circleci/config.yml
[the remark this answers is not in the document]
oferb: Ok

tools/build_or_test.sh
oferb: Cool :)

.circleci/config.yml
oferb: We have some python code on the way, and I wouldn't want to lose support for it.
Could you reinstate it?

.circleci/config.yml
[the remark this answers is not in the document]
oferb: :)
Let's call it `code`.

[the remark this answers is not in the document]
vmax: @oferb fixed in https://github.com/google/startup-os/commit/8bbd11a136c77a768a922fab21b6a9f544cf6ed1

.circleci/config.yml
vmax: @oferb fixed in https://github.com/google/startup-os/commit/8bbd11a136c77a768a922fab21b6a9f544cf6ed1

PR #568, opened 2019-04-20
Implementation of TypeScript formatter

vmax:
Adds support for formatting TypeScript files via [`tsfmt`]


vmax:
@oferb we need to merge hasadna/hasadna#183 first

PR #571, opened 2019-04-25
Sample gRPC server with SSL+FirebaseIdToken authentication

vmax:


common/grpc_auth/ClientAuthInterceptor.java
```diff
@@ -0,0 +1,27 @@
+package com.google.startupos.common.grpc_auth;
+
+import io.grpc.*;
```
oferb: Please import specific classes you need.

PR #575, opened 2019-04-29
Add findFilesByGlobPattern() method to FileUtils class

val-fed:
Add a method for getting file names in the path by pattern. The pattern can contain `*` and `**` wildcards.

The main goal of this method at the moment is the ability to use it for getting files defined by `glob` function - https://docs.bazel.build/versions/master/be/functions.html#glob

### DIFF
--- a/common/FileUtils.java
+++ b/common/FileUtils.java
@@ -37,12 +37,17 @@ import java.net.URL;
 import java.nio.file.FileSystem;
 import java.nio.file.FileSystems;
 import java.nio.file.FileVisitResult;
+import java.nio.file.FileVisitor;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.PathMatcher;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.StandardCopyOption;
 import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


common/FileUtils.java
```diff
@@ -472,5 +478,65 @@ public Message readPrototxtFromZipUnchecked(
...
+   * separated by the directory separator /. Throws IllegalStateException if path is not a folder.
+   */
+  public ImmutableList<String> findFilesByGlobPattern(String path, String pattern)
+      throws IOException {
+    if (!folderExists(path)) {
+      throw new IllegalStateException(String.format("%s folder does not exist", path));
+    }
+    Set<String> result = new HashSet<>();
+    final String filenamePattern = pattern.substring(pattern.lastIndexOf('/') + 1);
+    final PathMatcher matcher = fileSystem.getPathMatcher("glob:" + filenamePattern);
```
oferb: I'm looking at the options for glob here:
https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystem.html#getPathMatcher-java.lang.String-

Can you explain the case that is not covered there, for which you need to write custom code? Specifically, I mean not using whole `pattern` and only `filenamePattern`, and also needing `areFoldersMatchingPattern`.

common/FileUtils.java
```diff
@@ -472,5 +478,65 @@ public Message readPrototxtFromZipUnchecked(
   public File getFile(String path) {
     return new File(expandHomeDirectory(path));
   }
+
+  /**
+   * Gets file names in path by pattern. The pattern can containing the * wildcard: this wildcard
```
oferb: If you're using the standard glob format, you can also link to:
https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystem.html#getPathMatcher-java.lang.String-

